PR #181, opened 2014-11-06
separated exception classes into their own files

jesavro:
php-webdriver through composer and phing are not compatible since both declare NullPointerException classes and neither uses namespaces, cause a class redeclare error. This is a workaround until either project uses namespaces. In the meantime the composer autoloader will now pick either class but not both.


bertramakers:
I have the same problem (as I mentioned in #204).

Seeing how @whhone does not want to use namespaces, and excluding double classes is not an acceptable solution, I decided that renaming all Exception classes in php-webdriver so they end in "WebDriverException" could be a good compromise. My changes are in #205.

But as @jesavro mentioned, namespaces would really be the better long-term solution. The link mentioned by @whhone (https://jtreminio.com/2012/10/composer-namespaces-in-5-minutes/) does not fix the issue, it just explains how namespacing works. Composer can not "automagically" put your code in namespaces for you.

whhone:
I might take both pull requests 181 and 204.
Separating classes into separated files is a good practice and it makes sense to add WebDriverException given we don't use namespace.

gfosco:
Closing this as it is part of #209

PR #186, opened 2014-11-21
Do not hardcode CURL's request timeout, added ability to set custom timeout

OndraM:
Commit https://github.com/facebook/php-webdriver/commit/f666d1d41aebc9b663ed20a2ffe748d9a34a29f5 hardcoded curl's request request timeout to 300 seconds. 

But if you use Selenium in grid mode and have a long queue of waiting requests, it could take much longer than 5 minutes until the session create request is finished on the Selenium hub and the sessionId is acquired.  So you have to overwrite the timeout to something more reasonable than 300 seconds, what is now impossible in current master.

### DIFF
--- a/lib/remote/HttpCommandExecutor.php
+++ b/lib/remote/HttpCommandExecutor.php
@@ -133,24 +133,37 @@ class HttpCommandExecutor implements WebDriverCommandExecutor {
         'Accept: application/json',
       )
     );
-    curl_setopt($this->curl, CURLOPT_TIMEOUT_MS, 300000);
+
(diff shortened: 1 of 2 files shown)


OndraM:
Ping - what is your opinion, @whhone ?

whhone:
@OndraM see the current version of RemoteWebDriver here. https://github.com/facebook/php-webdriver/blob/master/lib/remote/RemoteWebDriver.php#L56

You can set both the request timeout and the connect timeout respectively.

OndraM:
Great, thanks.

PR #194, opened 2015-01-07
RemoteWebDriver: Allow specifying request timeout

danielbeardsley:
Previously you could only set the request timeout after the first
request (NEW_SESSION) was sent. If you want to alter the timeout, you
presumably want to affect the first request too.

### DIFF
--- a/lib/chrome/ChromeDriver.php
+++ b/lib/chrome/ChromeDriver.php
@@ -47,7 +47,8 @@ class ChromeDriver extends RemoteWebDriver {
   public static function create(
     $url = 'http://localhost:4444/wd/hub',
     $desired_capabilities = null,
-    $timeout_in_ms = 300000
+    $timeout_in_ms = 300000,
+    $request_timeout_in_ms = 300000
   ) {
     throw new WebDriverException('Please use ChromeDriver::start() instead.');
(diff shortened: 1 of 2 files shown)


whhone:
The operation like `$driver->get()` might take a long time but the first new session request should not exceed the time limit. I am feeling sick to add too much arguments to the constructor. 

Is there a strong reason to make this change? @danielbeardsley

danielbeardsley:
The reason is both practical and ideological. Both timeouts (connection and request) are allowed to be modified if need be, but because of the order of operations, altering the timeout for the first request wasn't possible. Practically, Selenium Grid can cause the initial `NEW_SESSION` request to take an arbitrary amount of time because they've implemented a queue, so it can be useful to alter the timeout.

The direct reason for adding this option is that in our usage of selenium we are getting odd `Read timed out` errors down deep in the guts of Selenium Grid. These cause a `NEW_SESSION` command to intermittently hang for exactly 150 seconds even though our timeouts for (further requests) are much shorter. Retrying the command after a hang often works, so we'd like the initial `NEW_SESSION` command to time out sooner than 5 minutes, but before this pull, that timeout wasn't adjustable.

PR #200, opened 2015-01-28
Add hasClass method

jameshd:
This proved useful for us. Not sure it's implemented correctly, and I've royally screwed up this PR / Forking business. Sorry.


whhone:
Thanks for the pull request. However, we would like to keep the library standardized as other language bindings. I suggest to create a class extending `RemoteWebDriver` or create a separated utility class for the `hasClass()`.

PR #201, opened 2015-02-08
Fixed strpos: ending parentheses is in incorrect location

arianf:
Fixed the current implementation of strpos, it currently looked like this

 `strpos($to_escape, "'" != false))` incorrectly passes `"'" != false` as the second argument.

Changed it to `if (strpos($to_escape, '"') !== false && strpos($to_escape, "'") !== false) {`


whhone:
Thanks! Nice catch!

PR #205, opened 2015-03-03
Resolved conflicts with other packages like Phing regarding Exception classes.

bertramakers:
These changes fix any conflicts of Exception class names with other packages like Phing. #181 attempted to fix this by putting the Exception classes in their own files, this pull request fixes the issue by renaming the Exception class so they all end in "WebDriverException" which makes it more clear what caused the Exception.

I realize that these class names were taken from the [the official Java API](http://selenium.googlecode.com/svn/trunk/docs/api/java/index.html?overview-summary.html), but keep in mind that the official Java API uses namespaces (packages) to prevent class name conflicts. As this project does not use namespaces, conflicts have to be reduced by making the class names as unique as possible.

### DIFF
--- a/lib/WebDriverExceptions.php
+++ b/lib/WebDriverExceptions.php
@@ -52,113 +52,113 @@ class WebDriverException extends Exception {
         // Success
         break;
       case 1:
-        throw new IndexOutOfBoundsException($message, $results);
+        throw new IndexOutOfBoundsWebDriverException($message, $results);
       case 2:
-        throw new NoCollectionException($message, $results);
+        throw new NoCollectionWebDriverException($message, $results);
       case 3:
-        throw new NoStringException($message, $results);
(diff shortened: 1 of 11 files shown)


bertramakers:
I realize that these class names were taken from the [the official Java API](http://selenium.googlecode.com/svn/trunk/docs/api/java/index.html?overview-summary.html), but keep in mind that the official Java API uses namespaces (packages) to prevent class name conflicts.

whhone:
I might take both pull requests. Separating classes into separated files is a good practice and it makes sense to add WebDriverException given we don't use namespapce.

bertramakers:
Taking both pull requests will result in merge conflicts, but seeing how #181 already has merge conflicts anyway I would suggest to merge this pull request in first and then fixing any old or new merge conflicts in #181.

bertramakers:
I created a combination of this PR and #181 in #209.

PR #210, opened 2015-04-20
add @access to docblocks

Tjoosten:
Add `@access` attribute to the docblocks

### DIFF
--- a/lib/JavaScriptExecutor.php
+++ b/lib/JavaScriptExecutor.php
@@ -23,8 +23,9 @@ interface JavaScriptExecutor {
    * of the currently selected frame. The executed script is assumed to be
    * synchronous and the result of evaluating the script will be returned.
    *
-   * @param string $script The script to inject.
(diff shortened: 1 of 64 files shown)


gfosco:
I'd allow adding the ~7 `@access private` lines, but the public ones seem superfluous.  If you want to update your PR with just that, I'll merge it.

Tjoosten:
How u mean srry not good at English

gfosco:
There are 7 places where you add:

```
@access private
```

I will allow adding those.  

Please remove all instances of `@access public`

Tjoosten:
so all the `@access public` needed to be removed?

gfosco:
Yeah, seems like it has no function.

0x20h:
Is `@access` necessary for PHP `>=5.3` anyway?
Seems a bit redundant.

gfosco:
True... Let's close this.

PR #217, opened 2015-05-27
Made RemoteWebElement extendable

drjayvee:
`newElement` uses the actual class' constructor to create new objects.

An alternative would have been to make the method protected (it is now private).

This change enables users to extend the `RemoteWebElement` API by subclassing it.


gfosco:
Sounds good.  Thanks!

PR #227, opened 2015-07-07
Fix: Update Composer binary in before_install section

localheinz:
This PR

* [x] updates Composer itself in the `before_install` section


gfosco:
Ok, seems useful.. Thanks.

PR #234, opened 2015-07-16
improved some @return type docs

drjayvee:
This makes my IDE happy :)


gfosco:
That makes me happy too :+1:

PR #237, opened 2015-07-23
Disable Firefox Reader View help tooltip

drjayvee:
This tooltip can hide window.document elements, which can break tests.
See http://stackoverflow.com/questions/30764805/how-to-disable-reader-view-in-firefox-using-webdriver

### DIFF
--- a/lib/Remote/DesiredCapabilities.php
+++ b/lib/Remote/DesiredCapabilities.php
@@ -195,10 +195,17 @@ class DesiredCapabilities implements WebDriverCapabilities {
    * @return DesiredCapabilities
    */
   public static function firefox() {
-    return new DesiredCapabilities(array(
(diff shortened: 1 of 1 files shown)


drjayvee:
Some clarification: Firefox 38.0.5 introduced a tooltip on the Reader View button in the address bar which pops up the very first time the button is shown. Since a fresh profile is used for every session, this tooltip is shown almost always.

This broke one of our tests because some elements were hidden behind that tooltip.

gfosco:
Looks okay to me...  Going to leave this open for another day or so.  Anyone else have a comment on whether this could cause other problems, or :+1: ?

drjayvee:
Hi! Just wanted to ask whether this will be merged.

gfosco:
:+1:

PR #243, opened 2015-08-03
unit tests wrong namespace deletion

EvanCoh:
As discussed here:
https://github.com/facebook/php-webdriver/issues/242


gfosco:
Going to leave this open for now, would like some community comment here if is the right thing to do or not.

PR #249, opened 2015-09-23
RemoteWebElement: allow overriding construction

danielbeardsley:
Previously, if a developer wanted to extend the RemoteWebElement()
class, you couldn't effectively do it (#217 got us part of the way).
The name of the class was hard-coded in a private factory
method of RemoteWebDriver that couldn't be overriden.

Now, a custom class that extends RemoteWebDriver can
construct a class derived from RemoteWebElement, allowing
extra functionality to be added to Elements themselves.

The same logic applies to the EventFiringWebElement.


gfosco:
Looks good to me.  Thanks :+1:

PR #256, opened 2015-10-22
FirefoxProfile: clean up temporary files

gkralik:
During profile setup, some temporary files/folders are created. These temporary files/folders can grow quite large, so clean them up after the profile has been constructed to prevent disk space issues.


gfosco:
What's the likelihood of these operations failing?  Could they be put in a try/catch/finally block to ensure they don't stop the process from continuing?

gkralik:
Both functions return `false` on failure and generate an `E_WARNING`,  so a failure would not halt processing.
But, thinking about it, `rmdir` will fail if the directory is non-empty. We need to recursively delete the directory and it's contents.

gfosco:
Ok, lets do it.  Thanks, :+1:

PR #271, opened 2015-12-21
timeout

jumper423:


gabrielpreston:
This is revisiting the same problem from https://github.com/facebook/php-webdriver/pull/198

I'm not sure this should be merged in, at least in its current form.

gfosco:
Cannot accept this PR to the community branch.  Invalid changes to composer.json (name, yii2), and we don't want to change the timeout.  Feel free to keep this in your own fork, of course.  Thanks.

PR #272, opened 2015-12-24
Use sys_get_temp_dir rather than the empty string when calling tempnam.

TobiasKappe:
The latter fails when using the empty string if an open_basedir restriction is
in effect.

Also refer to https://secure.php.net/manual/en/function.tempnam.php#93256


gfosco:
Cool, thanks for the contribution!

TobiasKappe:
Happy to help :)

PR #274, opened 2015-12-30
Improve unit tests & codestyle

OndraM:
Unit-tests were reorganized, so that their structure matches structure of the projects (what is usual in modern projects). Unit tests for `WebDriverCommand` and `DesiredCapabilities` was added.

Also some minor codestyle improvements (mainly missing phpdoc) were made.

Last but not least, class `FirefoxPreferences` with constants of common Firefox profile settings was added.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
   },
   "autoload-dev": {
     "psr-4": {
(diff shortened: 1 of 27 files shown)


gfosco:
Really great contribution.  Thanks! :+1:

PR #284, opened 2016-02-25
Enhance moveToElement method description

javigomez:
With the update in Selenium https://github.com/SeleniumHQ/selenium/issues/1202 the moveToElement function is going to take more relevance. I thought that the current description could lead to misunderstandings and decided to improve it a bit. I hope you find it more accurate.


javigomez:
just signed CLA

PR #298, opened 2016-05-09
Fixed PHP Fatal crash in RemoteWebDriver:507

sc0ttdav3y:
This PR fixes the issue noted at https://github.com/facebook/php-webdriver/issues/254

It's possible for 

    $this->executor

to be null at the time that this method is called, resulting in a PHP Fatal Error.

It is readily replicated within Codeception when an element cannot be found.  There's an issue over at Codeception https://github.com/Codeception/Codeception/issues/518 for this, but it's really a problem with RemoteWebDriver and not Codeception's consuming class. 

This PR adds a check for this condition and avoids the Fatal Error.

### DIFF
--- a/lib/Remote/RemoteWebDriver.php
+++ b/lib/Remote/RemoteWebDriver.php
@@ -504,7 +504,11 @@ class RemoteWebDriver implements WebDriver, JavaScriptExecutor {
       $params
     );
 
-    $response = $this->executor->execute($command);
-    return $response->getValue();
+	if ($this->executor) {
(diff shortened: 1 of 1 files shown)


facebook-github-bot-0:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

gfosco:
👍

PR #300, opened 2016-06-03
Throw exception if XPI extension file doesn't exist.

twindual:
Update to make extractTo() throw an exception if the Firefox extension .xpi file doesn't exist, which should be the desired behavior.

Currently extractTo() only throws an error if the .xpi file exists AND is unable to open.

ZipArchive throws some Warning messages when you use extractTo() and close() on an initialized ZipArchive object which is the case if the file doesn't exist.


facebook-github-bot-6:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

gfosco:
👍

Issue #301, opened 2016-06-04
Hardcoded 'em' namespace in installExtension() prevents some extensions from loading like uBlock Origin.


twindual:
Extensions that have a different namespace prefix than 'em' or don't use a namespace prefix for the 'id' element in the install manifest won't load.

The function installExtension() in the FirefoxProfiles.php file needs to be updated to avoid using a hard-coded namespace prefix.

The following example works because the 'id' element uses the hard-coded 'em' namespace.

'YouTube ALL HTML5' install.rdf file:
```
<?xml version="1.0" encoding="utf-8"?>
<RDF xmlns="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:em="http://www.mozilla.org/2004/em-rdf#">
    <Description about="urn:mozilla:install-manifest">
          <em:id>jid1-qj0w91o64N7Eeg@jetpack</em:id>
          <em:type>2</em:type>
          <em:bootstrap>true</em:bootstrap>
          <em:unpack>false</em:unpack>
          <em:version>3.0.3</em:version>
          <em:name>YouTube ALL HTML5</em:name>
```

The following 2 examples won't work because the 'id' element is not in the hard-coded 'em' namespace.

'Extension Options Menu' install.rdf file:
```
<?xml version="1.0"?>
<RDF:RDF xmlns:RDF="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns="http://www.mozilla.org/2004/em-rdf#">
	<RDF:Description about="urn:mozilla:install-manifest">
		<id>{1feca320-6b4d-11df-a08a-0800200c9a66}</id>
		<type>2</type>
		<name>Extension Options Menu</name>
```

'uBlock Origin' install.rdf file:
```
<?xml version="1.0" encoding="utf-8"?>
<r:RDF xmlns:r="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns="http://www.mozilla.org/2004/em-rdf#">
    <r:Description about="urn:mozilla:install-manifest">
        <id>uBlock0@raymondhill.net</id>
        <version>1.7.5</version>
        <name>uBlock Origin</name>
```

gfosco:
Merged #302 👍

PR #302, opened 2016-06-04
Fixed issue #301 - Hardcoded 'em' namespace in installExtension() prevents some extensions from loading

twindual:
Fixed installedExtension() in FirefoxProfile.php to remove hardcoded namespace prefix.

We now find the what namespace if any we need to use for the 'id' element before searching for it.


gfosco:
👍

Issue #212, opened 2015-05-13
Moving forward


gfosco:
In order to support both Facebook internal use and acceptance in the broader community, and considering the slow current slow pace of development (last release tagged 7 months ago,) we should maintain an official community version which is allowed to grow on its own.  

This way, the community can support current practices and interoperate better with other tools.  Important bug-fixes can be made to both versions, and everyone could help review and improve the code.  There are many forks, and some of that can be shared back for the benefit of all.

I've created a branch, community, which can become the default branch.  How about we use that to comply with PSR-2 and add Namespaces, per #204 and #190, and make a 1.0 release?

Any thoughts on this approach?

janoszen:
I have a working version with namespaces here: https://github.com/opsbears/php-webdriver

Saves you some work, just search and replace. :)

0x20h:
+1

OndraM:
+1, though this would mean rewriting our tests (or at least doing lots of search & replace).

janoszen:
@OndraM Please do check my fork, it's all done.

gabrielpreston:
I'm on board with this as well

0x20h:
@janoszen Why not prepare a PR?

janoszen:
Any requirements on the namespace part?

0x20h:
Just some thoughts:
- `Facebook\Webdriver` as root namespace (other proposals?)
- comply to PSR-4 (correct the autoload section in `composer.json`)
- Rename WebdriverException to `Facebook\Webdriver\Exception`
- Generally: follow one class per file
- put exceptions under `Facebook\Webdriver\Exception`

janoszen:
I can do that fairly quickly if that is ok.

OndraM:
@0x20h I would suggest keeping the name Web**D**river. It is also official name of the [W3C standard](http://www.w3.org/TR/webdriver/)..

0x20h:
@OndraM yep, absolutely. Thanks for the feedback!

gfosco:
I'm :+1: on all of those suggestions @0x20h @OndraM

janoszen:
OK guys, I have made the changes requested to my own repo (see above). It would need some real world testing, so if you have a project, feel free to jump on it. Feedback is also welcome.

0x20h:
Hey @janoszen, thanks for your work!
Just had a quick look at it. 
I think you should keep the directory structure and create sub namespaces.

janoszen:
[...]

Issue #190, opened 2014-12-05
Thoughts about coding style - PSR2?


OndraM:
Hi, as a following of PR #187, I'd like to ask about codingstyle of php-webdriver.

Current coding style is kind of weird, and does not resemble any other PHP coding style. Especially indenting (2 spaces are more usual for javascript and really unusual for PHP), variables $naming_in_underscores (while usual is $camelCase), brace placement and so on. Sometimes it is not even unified. 

This makes readability of the webdriver worse (during tests writing we often study the php-webdriver sourcecode to know how things work), also complicates contributing (almost nobody would have IDE set to current coding style).

I would suggest switching to [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md), as it is the most used and common coding style in current PHP world, it is default for various IDEs and most of new code in most used frameworks and libraries (Symfony, Zend Framework 2, Composer etc.) is in compliance with it, making it de-facto industry standard.

With tools like [code style fixer](http://cs.sensiolabs.org/) the transition is really easy and if you would agree, I'd be glad to participate on it. 

What are your opinions?

Tjoosten:
+1

gfosco:
Closing.. Please see #212

Issue #244, opened 2015-08-15
Microsoft Edge support


OndraM:
Since Selenium 2.47.0 the new Microsoft Edge browser is supported by the Selenium (and also by the java and dotnet official language bindings), see: https://github.com/SeleniumHQ/selenium/commit/9ed9272445806a336adf7cd3679bd3e5db364ae4

So that the `MicrosoftEdge` browser should be supported also by the php-webdriver.

However, as I do not personally use Windows, my possible contribution to this is rather limited :-/.

PR #311, opened 2016-07-30
PSR-2 codestyle

OndraM:
Hi, as was discussed in #212 and #190, I've finally adjusted the codestyle to match PSR-2. 
This can eg. allow simpler contributing, as PSR-2 is currently standard in most modern PHP libraries.

I've added also some useful checks on top of PSR-2 based on Symfony codestyle.

Last but not least I've added the codestyle checks to the Travis build.

What do you think? All code should be backwards compatible, the changes are only in the codestyle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 composer.phar
 composer.lock
 vendor
+.php_cs.cache
 
 # generic files to ignore
 *.lock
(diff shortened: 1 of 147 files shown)


gfosco:
👍  It looks pretty good to me... I'll wait a couple days if anyone else wants to comment, then look to merge.

OndraM:
So how does it look :-)? I started work on #244 , but would be great to have this merged first to avoid conflicts.

gfosco:
💥  🎆  🍻

gfosco:
tagged 1.1.3

PR #315, opened 2016-08-18
Initial Microsoft Edge support

OndraM:
This adds basic Microsoft Edge support - ie. only when Edge is used as a node. It still does not support starting Microsoft WebDriver executable locally - as it is [supported by eg. the official Java bindings ](https://github.com/SeleniumHQ/selenium/blob/c10e8a955883f004452cdde18096d70738397788/java/client/src/org/openqa/selenium/edge/EdgeDriverService.java), because this would require implementing the `EdgeDriverService` (which would extend `DriverService`, similary as [`ChromeDriverService` does](https://github.com/facebook/php-webdriver/blob/b7186fb1bcfda956d237f59face250d06ef47253/lib/Chrome/ChromeDriverService.php)). So the issue  #244 is still partially valid. But as a Linux user this full support would be quite difficult for me to implement and especially test - help from Windows users would be much appreciated!

However, when Edge is running as node (either on cloud service like Sauce Labs or BrowserStack) or as part of your Selenium grid, you can now start it easily using these predefined constant and desired capabilities:

```
$driver = RemoteWebDriver::create($host, DesiredCapabilities::microsoftEdge());
```

I also removed constants of some browsers no longer supported by Selenium, as per the [official  java bindings](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/BrowserType.java).

### DIFF
--- a/lib/Remote/DesiredCapabilities.php
+++ b/lib/Remote/DesiredCapabilities.php
@@ -274,6 +274,17 @@ class DesiredCapabilities implements WebDriverCapabilities
     /**
      * @return DesiredCapabilities
      */
+    public static function microsoftEdge()
+    {
+        return new self(array(
+            WebDriverCapabilityType::BROWSER_NAME => WebDriverBrowserType::MICROSOFT_EDGE,
+            WebDriverCapabilityType::PLATFORM => WebDriverPlatform::WINDOWS,
+        ));
+    }
+
+    /**
+     * @return DesiredCapabilities
+     */
     public static function iphone()
     {
         return new self(array(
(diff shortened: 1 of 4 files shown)


lib/Remote/DesiredCapabilities.php
```diff
@@ -274,6 +274,17 @@ public static function internetExplorer()
     /**
      * @return DesiredCapabilities
      */
+    public static function microsoftEdge()
```
gfosco: it's lowercase here but propercase everywhere else.. that ok?

OndraM: I know, but the method name just follows camelCase notation :-).

OndraM:
@gfosco Hi, should I do anything more to get this merged :-)?

gfosco:
👍

PR #317, opened 2016-09-01

sirdagor:
Curl server problem


OndraM:
@sirdagor Hi, what exactly is the problem? You are disabling certificates validation - what should definitely not be something that is enabled by default.

Issue #324, opened 2016-09-18
Drop PHP 5.3 & PHP 5.4 support?


OndraM:
Hi,
what do you think about setting minimal required version for this library to PHP 5.5?

The 5.3 and 5.4 versions of PHP are deprecated for a long time now and supporting them blocks some possibilities PHP gave us in 5.4 and 5.5 respectively. For instance traits (required by https://github.com/php-mock/php-mock-phpunit, which itself is required to unit test eg. curl_ calls), short array syntax, `::class` for class name resolution and much more. Right now I can't send PR to fix #323, as the unit tests depend on those PHP 5.5 features from php-mock-phpunit library...

According to statistics from https://seld.be/notes/php-versions-stats-2016-1-edition, the combined usage of 5.3 & 5.4 was in May already less than 10 %, and it perhaps dropped a bit more since then.

The 5.5 version is also deprecated for a few weeks now, but there is IMHO currently no need to drop its support yet (and it is still relatively widespread).

What do you think?

gfosco:
Open to this if more from the community want it.

gabrielpreston:
I think its a good idea as well. If there are projects that require older versions of PHP, they can lock to an older version of the driver.

drjayvee:
I'm in favor!

OndraM:
The PR is prepared in https://github.com/facebook/php-webdriver/pull/327

PR #327, opened 2016-09-22
Require PHP 5.5+

OndraM:
Ref #324.

Any suggestions what else could be changed/dropped/improved thanks to PHP 5.5 in current code base? (Besides the `::class` and short array syntax)


OndraM:
Rebased to current upstream and ready for approval & merge!

Issue #285, opened 2016-02-29
Difficult to extend WebDriverBy with fixed class names


oehlschl:
Sorry, mistakenly posted early.

I need to extend \Facebook\WebDriver\WebDriverBy but am slightly limited by the fact that the static ::className(), ::cssSelector(), etc. methods instantiate a new self instead of a new static by hardcoding the class name.

https://github.com/facebook/php-webdriver/blob/community/lib/WebDriverBy.php#L57

I've shimmed this to return new static() in my child class, but I'd prefer not to overwrite them when they could be readily changed at the source. A quick look at the php-webdriver codebase shows that the new static() paradigm is already in use, so there should be no issues with PHP version requirements, etc.

https://github.com/facebook/php-webdriver/search?q=%22new+static%22

Also, I would submit a pull request changing all instances of "new WebDriverBy" with "new static" in this class, but I don't have a Facebook account and thus cannot sign the requisite developer agreement.

DGCarramona:
I will try to make the PR this weekend, if I have time, hoping it will not be refused.
But why do you want to extend WebDriverBy ? Doesn't it manage yet all kind of selectors ?

oehlschl:
Thanks for the response.

For what it's worth I would be surprised if the PR were denied. LSB and self/static in this context have been around since PHP 5.3, and they're used elsewhere in this repo, so this case seems more defect than deliberate. If the argument is that WebDriverBy is not intended to be extended, then it should explicitly be made final and call new self() for clarity; the current implementation is dated and ambiguous.

The above is true independent of use case, but since you asked, we're using WebDriverBy elements as part of a third party acceptance testing framework, and extending them would allow us to print more information about the element in the test logs. Because it's a third party framework, we're also limited in what we can do, but adding a few additional properties and a __toString() method is a clean and reasonable implementation.

0x20h:
@oehlschl Would https://github.com/facebook/php-webdriver/blob/community/lib/Support/Events/EventFiringWebDriver.php help you to print more info using a listener?

oehlschl:
I appreciate the suggestion, but the point of this filing this issue was to improve the codebase by updating a class that adhered to an old, pre-LSB standard. Had the calls to new WebDriverBy() been replaced with new static() (which is reasonable for non-final, public methods), they could have been returned child classes as desired.

I see that these calls have since been replaced with new self(), which is less than ideal but at least adhering to current development practices.

Closing the issue since it's no longer a priority.

[...]

PR #333, opened 2016-10-05
Utilize late static binding

sepehr:
I came across the issue when I was trying to extend `WebDriverBy` class, it uses a whole bunch of `new self(...)` statements which restricts the extension by other classes. The same is for a few other classes.  

This PR aims to fix the issue across all affected classes by utilizing late static binding (`new static`) instead. LSB is introduced in PHP 5.3 which meets project's PHP version requirements.

This also addresses #285.


OndraM:
Great contribution, thanks @sepehr!

Issue #323, opened 2016-09-18
Using Codeception with Webdriver module and Appium


Propeller-SV:
I'm using Codeception with Webdriver module and Appium.app on MacOS. When I try to fill the field with:

`$I->fillField('form input[name=fname]', 'Tester');`

I get an error:

`Facebook\WebDriver\Exception\WebDriverException] JSON decoding of remote response failed.
Error code: 4
The response: 'Parameters were incorrect. We wanted {"required":["value"]} and you sent ["value",":id"]'`

[Appium logs](https://gist.github.com/Propeller-SV/45398867e2492221e922727e45d972f8)

Can I do something with this?

OndraM:
Hi, there could be a bug in HttpCommandExecutor class - it replaces the params in the URL (this is where the `:id` came from) but it do not unset the param from the other params being sent...

OndraM:
I may identified the bug. The attempted fix is my branch:
https://github.com/OndraM/php-webdriver/commit/1b2f6d5524b7387836846125c4b492eb7e131483

The issue lies there: https://github.com/OndraM/php-webdriver/commit/1b2f6d5524b7387836846125c4b492eb7e131483#diff-1ee156e0c33356b17f4dc2b3faec69eeL277 
The parameters are unset only if the method is NOT post - I guess the condition was unintentionally negated there, but I'm not sure. Maybe anyone other knows?

Could you test the fix? Maybe manually apply the changeset to your local copy of php-webdriver (in vendor folder).

The other issues is I prepared unit tests for this, however I'm not able to make them working in PHP 5.3 (what is currently the lowest supported version of PHP for php-selenium), as they require php-mock-phpunit library. So I'm not sure I can send the PR right now :-/.

Propeller-SV:
@OndraM 
Hello, I made manually changes in HttpCommandExecutor.php and now I cant see the bug, test is running like expected. Thank you very much!

OndraM:
@Propeller-SV Fixed in 1.2.0 release, so you can install this version.

PR #338, opened 2016-10-11
Fixes incompatibility with Appium

OndraM:
Do not set URL placeholder params as part of POST body. This fixes incompatibility with Appium, caused by redundant params present in requests to Selenium server.

See #323.


OndraM:
(Needs to be merged only after #327, as it works only for PHP 5.5+).

PR #370, opened 2016-12-15
Adding URL related conditions

davefx:


OndraM:
Hi @davefx, great contribution! May I ask you for the following, before we can marge this new feature?
- squash all the commits into one
- I would prefer naming the methods in the same as in the Java library - see [ExpectedConditions.java#L101](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java#L101) (so the name would be `urlToBe()`)
- What do you think about adding one more method (to follow the java library as well) - `urlMatches` ([ExpectedConditions.java#L147](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/support/ui/ExpectedConditions.java#L147)), which will check for regular expression in the URL? Would be also quite handy imho.

lib/WebDriverExpectedCondition.php
```diff
@@ -78,6 +78,38 @@ function (WebDriver $driver) use ($title) {
     }
 
     /**
+     * An expectation for checking the URL of a page.
+     *
+     * @param string $url The expected URL, which must be an exact match.
+     * @return bool WebDriverExpectedCondition True when the URL matches,
```
OndraM: There is no need to break the line here, the line limit is 120 characters :)

lib/WebDriverExpectedCondition.php
```diff
@@ -78,6 +78,38 @@ function (WebDriver $driver) use ($title) {
...
+            function (WebDriver $driver) use ($url) {
+                return $url === $driver->getCurrentURL();
+            }
+        );
+    }
+
+    /**
+     * An expectation for checking substring of the URL of a page.
+     *
+     * @param string $url The expected substring of the URL
```
OndraM: maybe call it `urlSubstring` for better readability?

OndraM:
@davefx Sorry to bother, but will you find time to finish the changes? It would be great to get it to a new version of php-webdriver! 

However, if you can't, I can do the adjustments myself (and keep you the credit of course). Let me know please :).

OndraM:
Replaced by #390, thanks @davefx!

OndraM:
New conditions were released as part of php-webdriver 1.3.0! Thanks again for your contribution, @davefx .

Issue #375, opened 2016-12-21
Allow option to select by partial text in the selectByVisibleText method.


cryptton2004:
Pre-requisites:
- Have a form that has multiple option labels:
```
<select>
   <option ...>1234 test</option>
   <option ...>5678 test</option>
   <option ...>9012 test</option>
</select>
```

Using:
`$getText->selectByVisibleText("1234");`
will return no results, wheareas it could also return the partial matches by changing:
    `$xpath = './/option[normalize-space(.) = '.$this->escapeQuotes($text).']';`

with:
    `$xpath = '//option[contains(text(), '.$this->escapeQuotes($text).')]';`

OndraM:
@cryptton2004 Hi, nice idea, I would rather not change current behavior, but it could e.g. be added as a `selectByVisiblePartialText()` method, what do you think? Could you send a pull request?

OndraM:
This was released as part of version 1.3.0.

PR #376, opened 2016-12-21
#375 - Allow option to select by partial text in the selectByVisiblePartialText…

cryptton2004:
#375 - Allow option to select by partial text in the selectByVisiblePartialText method.


OndraM:
Replaced by #391 . Thanks @cryptton2004 for your contribution 👍

PR #386, opened 2017-01-07
Send code coverage reports to coveralls.io

OndraM:


OndraM:
Ugh, forgot to disable those comments. Should be fixed now :)

PR #394, opened 2017-01-10
fix php docs for getText

gimler:
The getText return the real value with whitespaces leading and trailing. See also https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol

### DIFF
--- a/lib/Remote/RemoteWebElement.php
+++ b/lib/Remote/RemoteWebElement.php
@@ -262,7 +262,7 @@ class RemoteWebElement implements WebDriverElement, WebDriverLocatable
 
     /**
      * Get the visible (i.e. not hidden by CSS) innerText of this element,
-     * including sub-elements, without any leading or trailing whitespace.
+     * including sub-elements.
      *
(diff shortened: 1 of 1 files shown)


OndraM:
Hi @gimler , thanks for your contribution! However, the Selenium behavior is to remove all extra whitespaces from the text (leading, trailing, multiple non breaking inside the text atd.) - see eg. working draft of W3C WebDriver specification for further info: https://w3c.github.io/webdriver/webdriver-spec.html#h-get-element-text

> The Get Element Text command intends to return an element’s text “as rendered”. This is equivalent to calling element.innerText.

I also added PR with tests validating this behavior.
- The HTML content with spaces is shown [here](https://github.com/facebook/php-webdriver/pull/396/files#diff-4ed071b5837c6ef6f1ce087e91c3b2d3R22 )
- The returned value from getText (without spaces) is validated [here](https://github.com/facebook/php-webdriver/pull/396/files#diff-99b203f879028b8fc47b2819b2781425R30).

PR #404, opened 2017-01-26
fix liddle typo

gimler:


OndraM:
@gimler Hi, thanks fort the PR! However, the "falsey" is intentional term, meaning "values being casted to false". This is eg. `false`, `0`, `[]`, `null` etc., so not a typo :).

PR #410, opened 2017-02-06
Added ext-zip to require section of composer.json

NinoSkopac:
While trying to run this example (https://github.com/facebook/php-webdriver/wiki/upload-a-file) I ran into a problem of ZipArchive class not being found. Fixed it by doing `sudo apt-get install php7.1-zip`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
   "require": {
     "php": "^5.5 || ~7.0",
     "symfony/process": "^2.8 || ^3.1",
-    "ext-curl": "*"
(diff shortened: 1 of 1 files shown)


OndraM:
@NinoSkopac Hi, thanks for the contribution! However, the json file is not valid - there is a quote missing. Could you please fix the commit?

Thanks :)

NinoSkopac:
Done. Sorry for the f up :P

OndraM:
@NinoSkopac No problem, thank you! 👍

PR #413, opened 2017-02-10
Run functional tests on SauceLabs

OndraM:
Thanks to [SauceLabs](https://saucelabs.com/open-source), we can run functional tests against their cloud of platforms and browsers, allowing us to make sure the php-webdriver library is compatible with them.

The tests results will be publicly available here: https://saucelabs.com/u/php-webdriver


gfosco:
👍 👍 💯

PR #436, opened 2017-04-07
Support Symfony/process v3.0

thejanasatan:


OndraM:
Hi @thejanasatan , thanks for the contribution!

However, the constrain change is not necessary to support symfony/process 3.1 (or later 3.x version) - the caret operator in `^3.0` means "3.0 or any later compatible versions", ie. 3.0.1, 3.1.0 or 3.2.0 etc. See also https://getcomposer.org/doc/articles/versions.md#caret

You can also check this change selects the same versions as the current constrain. Compare:
https://semver.mwl.be/#?package=symfony%2Fprocess&version=%5E2.8%20%7C%7C%20%5E3.0&minimum-stability=stable
vs
https://semver.mwl.be/#?package=symfony%2Fprocess&version=%5E2.8%20%7C%7C%20%5E3.0%20%7C%7C%20%5E3.1&minimum-stability=stable

PR #438, opened 2017-04-27
Correct constant checking

toniperic:
Prior to this change I would be getting `Constant CURLOPT_CONNECTTIMEOUT_MS already defined` since PHP would try to re-define a constant.


OndraM:
@toniperic 👍 Nice, thanks!

toniperic:
@OndraM you're welcome.

Out of curiosity: when do you tag new (patch) versions?

OndraM:
@toniperic I'm just going to release 1.4.1 with this fix in a few moments :).

toniperic:
@OndraM sounds great. Thanks! 👍

PR #447, opened 2017-06-21
Add LICENCE.md

OndraM:
The licence is already defined in header of all files (and is also stated on packagist). However adding the licence text directly to the repository is more standard, as Github will recognize it by itself, add it to the repository header etc.


OndraM:
@gfosco Hi, do you think this is good to merge?

PR #461, opened 2017-08-21
Update platform capability to match the w3c spec

paulbriton:
According to the [spec](https://w3c.github.io/webdriver/webdriver-spec.html#capabilities), the capability platform has changed : 

>"platformName" 
> Lowercase name of the current platform as a string.

### DIFF
--- a/lib/Remote/WebDriverCapabilityType.php
+++ b/lib/Remote/WebDriverCapabilityType.php
@@ -23,6 +23,7 @@ class WebDriverCapabilityType
     const BROWSER_NAME = 'browserName';
     const VERSION = 'version';
     const PLATFORM = 'platform';
+    const PLATFORM_NAME = 'platformName';
     const JAVASCRIPT_ENABLED = 'javascriptEnabled';
     const TAKES_SCREENSHOT = 'takesScreenshot';
     const HANDLES_ALERTS = 'handlesAlerts';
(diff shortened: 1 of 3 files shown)


paulbriton:
By the way, I think we could remove Windows XP since some drivers doesn't support it, as well as Microsoft..

OndraM:
Hi @paulbriton and thanks for the contribution!

FYI this pull request is now intentionally on hold. We cannot simply change features to support W3C protocol, as it could break current OSS JsonWire protocol (and won't be supported there either).

So this will be solved as part of #469 , where this will be part of W3C adapter (and will be untouched for JsonWire protocol).

So I'm keeping this PR open for now, will take care of it later.

gohelkiran:
any updates on this? there are many errors with UnknownCommandException with latest versions of Geckodriver, Firefox and Selenium web driver.
Any plan in near future or solutions to solve this errors
https://github.com/facebook/php-webdriver/issues/580

OndraM:
@dunglas , do you think this should be meged into #560 or done separately? Or this should be part of the Adapter layer?

dunglas:
@OndraM I've not looked at capabilities yet, I've no idea. If this change is compatible with Chrome, we may merge it separately?

OndraM:
This was superseded by https://github.com/facebook/php-webdriver/pull/678 - capabilities should be W3C compatible in upcoming release 1.8.0 :tada: .

Thanks @paulbriton and sorry for letting you wait :)

PR #473, opened 2017-09-08
When creating driver from existing session, retrieve capabilities from the browser

OndraM:
So the driver instance is in a same state as when new browser is created.


OndraM:
There is no such endpoint to retrieve current capabilitites in W3C WebDriver, meaning there is no reason to add this functionality just for the JsonWire.

PR #494, opened 2017-11-03
Adding WebDriverExpectedCondition::visibilityOfAnyElementsLocated

qheaden:
This pull request adds a new visibilityOfAnyElementsLocated method to WebDriverExpectedCondition. This condition checks to see if any elements in an array of located elements are visible, not just the first one. This condition also returns an array of all elements that meet the criteria and are located and visible.

### DIFF
--- a/lib/WebDriverExpectedCondition.php
+++ b/lib/WebDriverExpectedCondition.php
@@ -194,6 +194,35 @@ class WebDriverExpectedCondition
     }
 
     /**
+     * An expectation for checking than at least one element in an array of elements is present on the
+     * DOM of a page and visible.
+     * Visibility means that the element is not only displayed but also has a height and width that is greater than 0.
+     *
+     * @param WebDriverBy $by The located used to find the element.
+     * @return WebDriverExpectedCondition<WebDriverElement> Condition returns the elements that are located and visible.
+     */
+    public static function visibilityOfAnyElementLocated(WebDriverBy $by)
+    {
+        return new static(
+            function (WebDriver $driver) use ($by) {
+                $elements = $driver->findElements($by);
+                $visibleElements = [];
+
+                foreach ($elements as $element) {
+                    try {
+                        if ($element->isDisplayed()) {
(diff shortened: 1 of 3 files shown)


OndraM:
Hi @qheaden , thanks for the contribution, I love you also added unit test! 💯 

May I ask what is the use-case for this condition? We need to make sure this is not too specific scenario to be make this included as a part of the library :).

And second thing - could you please fixup all the commits into one?

qheaden:
Thanks for looking at the PR @OndraM. Here is an example use case. Suppose you have a web page that allows searching items, but on initial load puts all items on the page in a non-visible state. As you select different search criteria, items that match that criteria are put into a visible state, but the elements that become visible might not be first in the list of elements.

Because the code that implements visibilityOfElementLocated uses findElement instead of findElements, only the first element is checked for visibility. So in this case, if the 2nd, 3rd, or higher element because visible, the check will still fail because it will only check visibility of the first one.

I hope I have made sense. :smile:

lib/WebDriverExpectedCondition.php
```diff
@@ -194,6 +194,34 @@ function (WebDriver $driver) use ($by) {
...
+            function (WebDriver $driver) use ($by) {
+                $elements = $driver->findElements($by);
+                $visibleElements = [];
+
+                foreach ($elements as $element) {
+                    try {
+                        if ($element->isDisplayed()) {
+                            $visibleElements[] = $element;
+                        }
+                    } catch (StateElementReferenceException $e) {}
```
OndraM: This needs to be on two lines (see failed Travis build).
```
} catch (StateElementReferenceException $e) {
} 
```

OndraM:
@qheaden OK, makes sense to me. Please fix the one code-style issue and I'm gonna merge it! 🎉

qheaden:
Closing and re-opening to trigger rebuild.

OndraM:
@qheaden Merged, thanks! 🎉

PR #495, opened 2017-11-03
Fix code style

VolCh:
Fixed https://travis-ci.org/facebook/php-webdriver/jobs/296256023 for just released php-cs-fixer 2.8.0


OndraM:
Thanks again, @VolCh!

Reference: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3114 .

PR #498, opened 2017-11-04
Use PSR-1 for PHPUnit TestCase

gabriel-caruso:
Use [PSR-1](http://www.php-fig.org/psr/psr-1/#namespace-and-class-names) while extending PHPUnit TestCase class. This will help us when to migrate to PHPUnit 6, that [no longer support snake case namespaces](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     "ext-zip": "*"
   },
   "require-dev": {
-    "phpunit/phpunit": "^5.4",
+    "phpunit/phpunit": "^5.7",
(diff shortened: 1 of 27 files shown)


OndraM:
@gabriel-caruso Great, thanks for your first contribution 👍 . A also changed namespace in TestListener and rebases the tests. Will merge this right after Travis is done 🎉 .

OndraM:
Merged, thanks @gabriel-caruso!

Issue #513, opened 2017-12-04
Take element screenshot not implemented in php-webdriver


sapzape:
As per spec: https://w3c.github.io/webdriver/webdriver-spec.html#dfn-take-element-screenshot
Take element screenshot is not implemented in php-webdriver. It should use the following endpoint:

GET /session/{session id}/element/{element id}/screenshot

It provides a guide, but I think the method implementation is more efficient in usability.
(guide : https://github.com/facebook/php-webdriver/wiki/taking-full-screenshot-and-of-an-element)

If you agree with the need for implementation, can I implement it?

OndraM:
Implemented in #701 :tada:

Issue #469, opened 2017-09-01
W3C WebDriver protocol support


OndraM:
This is an issue to plan and track progress of implementing [W3C WebDriver](https://w3c.github.io/webdriver/webdriver-spec.html) protocol into php-webdriver.

Currently, only legacy [JsonWire](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol) (aka OSS) protocol is supported by php-webdriver. Selenium server 3.x translates the protocol if needed for the remote end (browser). However not everything is translated (like `actions()`), making some of the features not working with remote ends supporting only the W3C protocol (which is currently only GeckoDriver for Firefox 48+).

Since Selenium 3.5 the translation to old protocol is disabled by default. **It still could be reenabled** by starting the server like `java -jar selenium-server-standalone-3.5.3.jar -enablePassThrough false`, but this is only partial and temporary solution/workaround.

For Docker 🐳  users: use SE_OPTS environment variable to pass the enablePassThrough option to Selenium server. Ie. add `-e "SE_OPTS=-enablePassThrough false"` to `docker run` command.

## The work would require eg.:
- add some more functional tests, to make sure we don't breake the old protocol while doing the necessary refactoring
- prepare handshake to detect whether the remote end is speaking JsonWire or W3C dialect (we need this to support the old browsers etc.)
- define common interface for Json and W3C protocol
- convert current implementation to use this interface
- implement adapter for the W3C protocol
- implement GeckodriverService or something like that to start geckodriver locally
...

## Resources
- [JsonWireProtocol vs. W3C WebDriver](https://github.com/facebook/php-webdriver/wiki/JsonWireProtocol-vs.-W3C-WebDriver)
- [ProtocolHandshake](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/ProtocolHandshake.java) in java bindings
- [Handshake and capabilities resolving](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webdriver.py#L252) in python bindings

OndraM:
I started working on https://github.com/facebook/php-webdriver/wiki/JsonWireProtocol-vs.-W3C-WebDriver . I'd like to summarize there the key protocol differences before we start actually implementing them in the code. 

Any help with this overview document will be much appreciated!

paulbriton:
Spec also contains changes about the JSON returned when finding an element. 
[...]

PR #515, opened 2017-12-06
Implemented TakeElementScreenshot

sapzape:
This PR add a new method takes a screenshot of specific element.

ISSUE=#513


OndraM:
Hi @sapzape and thanks for the contribution!

FYI, this pull request is now intentionally on hold. We cannot simply add features specific for W3C protocol, as it could break current OSS JsonWire protocol (and won't be supported there either).

So this will be solved as part of #469 , where this method for partial screenshot will be part of W3C adapter.

So I'm keeping this PR open for now, will take care of it later.

lib/Remote/HttpCommandExecutor.php
```diff
@@ -123,6 +123,7 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
         ],
         DriverCommand::SUBMIT_ELEMENT => ['method' => 'POST', 'url' => '/session/:sessionId/element/:id/submit'],
         DriverCommand::SCREENSHOT => ['method' => 'GET', 'url' => '/session/:sessionId/screenshot'],
+        DriverCommand::ELEMENT_SCREENSHOT => ['method' => 'GET', 'url' => '/session/:sessionId/element/:id/screenshot'],
```
andrewnicols: This is a W3C only feature .

lib/Remote/DriverCommand.php
```diff
@@ -76,6 +76,7 @@ class DriverCommand
     const GET_ELEMENT_VALUE_OF_CSS_PROPERTY = 'getElementValueOfCssProperty';
     const ELEMENT_EQUALS = 'elementEquals';
     const SCREENSHOT = 'screenshot';
+    const ELEMENT_SCREENSHOT = 'elementScreenshot';
```
andrewnicols: The W3C spec talls this "Take Element Screenshot" so I'd suggest we go for TAKE_ELEMENT_SCREENSHOT and takeElementScreenshot to match.

lib/Remote/RemoteWebElement.php
```diff
@@ -403,6 +403,27 @@ public function equals(WebDriverElement $other)
         ]);
     }
 
+    /**
+     * Take screenshot of a specific element.
+     *
+     * @param string $save_as The path of the screenshot to be saved.
+     * @return string The screenshot in PNG format.
+     */
+    public function takeElementScreenshot($save_as = null)
```
andrewnicols: camelCale $saveAs

sapzape:
I'll update again when it's decided how.

lib/Remote/DriverCommand.php
OndraM: Agreed, also after you rebase the pull request, please move this constant to the end of the file (we now have there `// W3C specific`).

lib/Remote/HttpCommandExecutor.php
OndraM: But it can stay here, right? The `$w3cCompliantCommands` array below in the file (in current community branch) is used for overloading methods which have different params in W3C. Methods which are the same could stay here.

lib/Remote/RemoteWebElement.php
OndraM: We are now mixing camelCase and underscore_case in the codebase (will be united to camelCase in 2.0), so anything is now acceptable :)

lib/WebDriverElement.php
```diff
@@ -132,4 +132,12 @@ public function submit();
...
      */
     public function getID();
+
+    /**
+     * Take screenshot of a specific element.
+     *
+     * @param string $save_as The path of the screenshot to be saved.
+     * @return string The screenshot in PNG format.
+     */
+    public function takeElementScreenshot($save_as = null);
```
OndraM: Extending interface is something we unfortunately can't do in 1.x - because if would break backwards compatibility.

So please comment the line and add some todo like this:

```
    /**
     * Take screenshot of a specific element.
     *
     * @param string $save_as The path of the screenshot to be saved.
     * @return string The screenshot in PNG format.
     * @todo Add in next major release (BC)
     */
    //public function takeElementScreenshot($save_as = null);
```
We will then uncomment it in 2.0.

OndraM:
Ping @sapzape - are you able to finish the PR? Would be great! (Sorry for pushing this now - but we are preparing for release 1.8 and would like to make this part of it :pray: ).

Thanks :100:

lib/Remote/HttpCommandExecutor.php
sapzape: I kept the existing code.

lib/Remote/DriverCommand.php
sapzape: Done

lib/Remote/RemoteWebElement.php
sapzape: I kept the existing code.

lib/WebDriverElement.php
sapzape: Done

sapzape:
All requests have been updated.
@OndraM  Should I write test code in this patch? (cc. @andrewnicols )

lib/Remote/HttpCommandExecutor.php
andrewnicols: But you're still referencing ELEMENT_SCREENSHOT when the constant has been renamed to TAKE_ELEMENT_SCREENSHOT.

andrewnicols:
Hi @sapzape ,

Unit tests are always good to have if you're able to provide them.

A couple of other points:
1. Incorrect constant use in the place I've already commented
1. You've also included both the PSR2 commit, and a merge commit.

You should be able to drop both the PSR2 and merge commit, and then squash your final commit into the original one. This then looks good from my side.

sapzape:
The rebase is invalid and so i will create a new PR.

PR #521, opened 2017-12-29
Add `ext-mbstring` to requirements

alexmnv:
`mb_strpos` function is used in several places in the project. Many features will not work without `ext-mbstring` installed.


OndraM:
Hi @alexmnv, you are right - thanks for the contribution!

PR #528, opened 2018-02-02
Update PHPDoc for functions that return static instances of a class

dbudwin:
Previously, functions that contained functionality similar to `return new static(...)` had their `@return` PHPDoc tag to specify the class name like `@return SomeClass` where the preferred documentation should be `@return static`.

Some support for this decision:
https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#keyword
https://netbeans.org/bugzilla/show_bug.cgi?id=196565
https://blog.madewithlove.be/post/return-types/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This project versioning adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 ### Added
 - Connection and request timeouts could be specified also when creating RemoteWebDriver from existing session ID.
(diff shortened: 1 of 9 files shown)


OndraM:
Hi @dbudwin , thanks for the contribution! You are right, the `static` keyword in phpDoc should be used there. I rebased you branch and I will merge it as soon as the Travis builds are finished.

👍 Thanks again!

dbudwin:
Awesome, thanks!  Looking forward to it, let me know if I can be of further assistance.

PR #529, opened 2018-02-08
Add .vscode folder to .gitignore

Stephanyan:
Visual Studio Code users often generate a .vscode folder that contains a "workspace settings" file (settings.json) that overrides the "User settings" only for a specific project, exactly like the .idea folder does.

Like "[.idea](https://github.com/facebook/php-webdriver/blob/community/.gitignore)", this folder should be ignored by Git or any version control system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ logs/
 *~
 *.swp
 .idea
+.vscode


OndraM:
Generally, you should use global gitignore on your system, which will include those files, depending on IDE and operating system you use. But as this repo is often used by people without strong developer background, this may be easier for them. 

Thanks for the contribution 👍

PR #530, opened 2018-02-12
add reply request without 100-continue expectation

antoninrykalsky:
This pull-request solves strict RFC HTTP/1.1 https://tools.ietf.org/html/rfc7231#section-5.1.1 behavior.

If you are communicating with a server on HTTP/1.0 old behavior is ok. But if HTTP/1.1 is used, by RFC HTTP 417 should be returned. This will probably will work on most cases, but in combination with squid proxy, you will be forced to solve it on app site.

HTTP Expect: 100-continue is added when `curl_setopt($this->curl, CURLOPT_POSTFIELDS, $encoded_params);` with not null `$encoded_params` present.

### DIFF
--- a/lib/Remote/HttpCommandExecutor.php
+++ b/lib/Remote/HttpCommandExecutor.php
@@ -263,6 +263,12 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
             curl_setopt($this->curl, CURLOPT_CUSTOMREQUEST, $http_method);
         }
 
+        if (in_array($http_method, ['POST', 'PUT'])) {
+            // Disable sending 'Expect: 100-Continue' header, as it is causing issues with eg. squid proxy
+            // https://tools.ietf.org/html/rfc7231#section-5.1.1
+            curl_setopt($this->curl, CURLOPT_HTTPHEADER, ['Expect:']);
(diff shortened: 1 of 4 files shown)


lib/Remote/HttpCommandExecutor.php
```diff
@@ -273,6 +273,16 @@ public function execute(WebDriverCommand $command)
 
         $raw_results = trim(curl_exec($this->curl));
 
+        # strict RFC http/1.1 https://tools.ietf.org/html/rfc7231#section-5.1.1
```
OndraM: Please use standard multiline comments with `/* [multiple lines] */`

antoninrykalsky:
@OndraM Right, I guess selenium servers will always implement HTTP/1.1 or newer so that 100-continue can be removed. I have done it for all POST/PUT as suggested.

OndraM:
@antoninrykalsky Thanks! Though some unit tests are now broken and need to be fixed (just change the values in mock expectations). I may do it during tomorrow, or you may do it yourselves if you are interested or want it sooner :-).

OndraM:
Merged. Thanks @antoninrykalsky! 👍

PR #533, opened 2018-02-22
Update README to warn about NOT updating to 3.9

lgraziani2712:
Selenium added a breaking change and the driver won't work.

See #469


OndraM:
Hi @lgraziani2712 and thanks for the contribution. However, it is based on my inaccurate statement. I updated it a little bit in the referenced issue to make this clear, so I'm gonna close this PR.

Thanks anyway!

PR #535, opened 2018-03-05
Minor tweak to environment variable name.

AtricoSoftware:
Allow a second attempt to read a bash compatible valriable

1. Why is this change necessary?
Current environment variable in ChromeDriver.php uses '.', whilst valid according to the specification, this will not work using a bash shell (probably the most commonly used)

2. How does it address the issue?
'.' replaced with '_'.  This is usable in bash (including cygwin)
Original code still there, only attempts new variable if original does not exist

3. What are the side effects?
None, previous name will still work, new name only used if original name does not exist


AtricoSoftware:
Hmm, strange test failure (in testing with MS edge, I modified only ChromeDriver code)
Looking at test output, I can't see how my changes could have broken THIS test without breaking many others.
Is this branch broken?

OndraM:
Replaced by #696, thanks @AtricoSoftware!

PR #544, opened 2018-03-13
Get back other http headers (Content type, Accept) in case of adding …

antoninrykalsky:
A colleague has found that this pull request brokes communication with Appium. Appium response gives `Invalid parameter response`, in the case when Except header is set. 

After some investigation, I have found that my code
`curl_setopt($this->curl, CURLOPT_HTTPHEADER, ['Expect:'];`

do not add the header, but re-sets all HTTP headers. So it removes those from constructor part:
`[
                'Content-Type: application/json;charset=UTF-8',
                'Accept: application/json'
]`

### DIFF
--- a/lib/Remote/HttpCommandExecutor.php
+++ b/lib/Remote/HttpCommandExecutor.php
@@ -26,6 +26,11 @@ use InvalidArgumentException;
  */
 class HttpCommandExecutor implements WebDriverCommandExecutor
 {
+    const DEFAULT_HTTP_HEADERS = [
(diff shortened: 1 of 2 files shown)


OndraM:
Hi @antoninrykalsky , I updated the PR a bit (this way it would mean adding another Expect header each time the Executor is called to execute some command, because the curl instance is shared for the script lifetime). When Travis is done I will merge it. Thanks!

antoninrykalsky:
Hi @OndraM. Good point, thanks

PR #545, opened 2018-03-13
Add utility classes to interact with checkboxes and radio buttons

### LINKED ISSUE #373: Convenience class for checkbox
Similary as we have [WebDriverSelect](https://github.com/facebook/php-webdriver/blob/community/lib/WebDriverSelect.php) for `<select>`, it may be handy to have also something like WebDriverCheckbox for `<input type="checkbox">` elements with helper methods:
- check
- uncheck
- isChecked

and maybe some more.

dunglas:
This PR adds an utility class to easily manipulate checkboxes and radio buttons.

Imagine the following form:

```html
<form id="my-form" method="POST">
    <input type="checkbox" name="c" value="1">

    <label>
        Checkbox Foo

        <input type="checkbox" name="c" value="2">
    </label>

    <input type="radio" name="r" value="x">

    <label>
        Radio Foo

        <input type="radio" name="r" value="y">
    </label>

    <input type="submit" value="OK">
</form>

<label for="c1">Checkbox Bar</label>
<input type="checkbox" form="my-form" name="c" value="3" id="c1">

<label for="r1">Radio Bar</label>
<input type="radio" form="my-form" name="r" value="z" id="r1">
``` 

The proposed class allows to interact with it in a smooth way:

```php
$checkbox = new WebDriverCheckbox($this->driver->findElement(WebDriverBy::xpath('//input[@type="checkbox"]'))); // select any checkbox
$radio = new WebDriverRadio($this->driver->findElement(WebDriverBy::xpath('//input[@type="radio"]'))); // Select any radio button

var_dump($checkbox->getOptions()); // 1, 2, 3
var_dump($radio->getOptions()); // x, y, z

$checkbox->selectByValue('1');
$checkbox->selectByVisibleText('Checkbox Foo');
$checkbox->selectByVisiblePartialText('Bar');
$checkbox->deselectByVisiblePartialText('Bar');

$radio->selectByValue('a');
$radio->selectByIndex(1);

var_dump($checkbox->getAllSelectedOptions()); // 1, 2
var_dump($radio->getAllSelectedOptions()); // b

var_dump($radio->getFirstSelectedOption()); // b

$checkbox->deselectAll());
```

Features:

* Checkbox and radio buttons support
* Ability to select an element by its label (full or partial matching)
* Ability to select an element by its (pseudo) 0-based index

For consistency and convenience - especially when using libraries such as Symfony Form that allow to switch from checkboxes to selects in 1 line - the new class mimics the public API of `WebDriverSelect` and implements the `WebDriverSelectInterface`. 

Closes #373.

### DIFF
--- a/lib/WebDriverCheckbox.php
+++ b/lib/WebDriverCheckbox.php
@@ -1,0 +1,288 @@
+<?php
+// Copyright 2004-present Facebook. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Facebook\WebDriver;
+
+use Facebook\WebDriver\Exception\NoSuchElementException;
+use Facebook\WebDriver\Exception\UnexpectedTagNameException;
+use Facebook\WebDriver\Exception\UnsupportedOperationException;
+use Facebook\WebDriver\Exception\WebDriverException;
+use Facebook\WebDriver\Support\XPathEscaper;
+
+/**
+ * Provides helper methods for checkboxes and radio buttons.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class WebDriverCheckbox implements WebDriverSelectInterface
+{
+    private $element;
+    private $type;
+    private $name;
+
+    public function __construct(WebDriverElement $element)
+    {
+        if ('input' !== $tagName = $element->getTagName()) {
+            throw new UnexpectedTagNameException('input', $tagName);
+        }
+
+        $type = $element->getAttribute('type');
+        if ('checkbox' !== $type && 'radio' !== $type) {
+            throw new WebDriverException('The input must be of type "checkbox" or "radio".');
+        }
+
+        if (null === $name = $element->getAttribute('name')) {
+            throw new WebDriverException('The input have a "name" attribute.');
+        }
+
+        $this->element = $element;
+        $this->type = $type;
+        $this->name = $name;
+    }
+
+    public function isMultiple()
+    {
+        return 'checkbox' === $this->type;
+    }
+
+    public function getOptions()
+    {
+        return $this->getRelatedElements();
+    }
+
+    public function getAllSelectedOptions()
+    {
+        $selectedOptions = [];
+        foreach ($this->getRelatedElements() as $element) {
+            if ($element->isSelected()) {
+                $selectedOptions[] = $element;
+
+                if (!$this->isMultiple()) {
+                    return $selectedOptions;
+                }
+            }
+        }
+
+        return $selectedOptions;
+    }
+
+    public function getFirstSelectedOption()
+    {
+        foreach ($this->getRelatedElements() as $element) {
+            if ($element->isSelected()) {
+                return $element;
+            }
+        }
+
+        throw new NoSuchElementException('No options are selected');
+    }
+
+    public function selectByIndex($index)
+    {
+        $this->byIndex($index);
+    }
+
+    public function selectByValue($value)
+    {
+        $this->byValue($value);
+    }
+
+    public function selectByVisibleText($text)
+    {
+        $this->byVisibleText($text);
+    }
+
+    public function selectByVisiblePartialText($text)
+    {
+        $this->byVisibleText($text, true);
+    }
+
+    public function deselectAll()
+    {
+        if (!$this->isMultiple()) {
+            throw new UnsupportedOperationException('You may only deselect all options of checkboxes');
+        }
+
+        foreach ($this->getRelatedElements() as $option) {
+            $this->deselectOption($option);
+        }
+    }
+
+    public function deselectByIndex($index)
+    {
+        if (!$this->isMultiple()) {
+            throw new UnsupportedOperationException('You may only deselect checkboxes');
+        }
+
+        $this->byIndex($index, false);
+    }
+
+    public function deselectByValue($value)
+    {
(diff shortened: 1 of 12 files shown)


lib/WebDriverCheckbox.php
```diff
@@ -0,0 +1,290 @@
...
+
+            try {
+                $element->findElement(WebDriverBy::xpath($xpathNormalize));
+            } catch (NoSuchElementException $e) {
+                if ($partial) {
+                    continue;
+                }
+
+                try {
+                    // Since the mechanism of getting the text in xpath is not the same as
```
dunglas: I'm not sure that this case really occurs. But as the `WebDriverSelect` already does that, I've kept this behavior.

lib/WebDriverCheckbox.php
```diff
@@ -0,0 +1,296 @@
...
+use Facebook\WebDriver\Exception\UnsupportedOperationException;
+use Facebook\WebDriver\Exception\WebDriverException;
+use Facebook\WebDriver\Support\XPathEscaper;
+
+/**
+ * Provides helper methods for checkboxes and radio buttons.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class WebDriverCheckbox implements WebDriverSelectInterface
```
OndraM: I suggest splitting the class in two (Checkbox and Radio), because:
- the semantic is confusing (`$radiobutton = new WebDriverCheckbox($radibuttonElement)`)
- there are lot of `$this->isMultiple()` conditionals, which separates logic for checkboxes and radiobuttons

Maybe some common abstract class could be introduced to DRY the logics. What do you think?

tests/functional/web/form_checkbox_radio.html
```diff
@@ -0,0 +1,35 @@
...
+    <meta charset="UTF-8">
+    <title>Form</title>
+</head>
+<body>
+<form id="another-form" method="POST">
+    <input name="j1">
+    <input type="checkbox" name="j2" value="j2a">
+
+    <label>
+        J2B
```
OndraM: To make sure the normalize-space() selectors work as expected, there could also be a case where the visible text has multiple spaces inside, ie. not being just one word, like:
 <pre>
<label>
   J     2 

  B    
...
 </pre>

What are you thoughts?

lib/WebDriverCheckbox.php
```diff
@@ -0,0 +1,296 @@
...
+    {
+        return 'checkbox' === $this->type;
+    }
+
+    public function getOptions()
+    {
+        return $this->getRelatedElements();
+    }
+
+    public function getAllSelectedOptions()
```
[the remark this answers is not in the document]
dunglas: I definitely agree, the interface is too generic, but it's convenient to have a common one. Introducing the new one can even be done (in another PR?) without breaking changes.

lib/WebDriverCheckbox.php
```diff
@@ -0,0 +1,296 @@
...
+            $xpathNormalize = sprintf('%s[%s]', $xpath, $normalizeFilter);
+            if (null !== $id = $element->getAttribute('id')) {
+                $idFilter = sprintf('@for = %s', XPathEscaper::escapeQuotes($id));
+
+                $xpath .= sprintf(' | //label[%s]', $idFilter);
+                $xpathNormalize .= sprintf(' | //label[%s and %s]', $idFilter, $normalizeFilter);
+            }
+
+            try {
+                $element->findElement(WebDriverBy::xpath($xpathNormalize));
```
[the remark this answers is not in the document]
dunglas: See https://github.com/facebook/php-webdriver/pull/545#discussion_r174204860
It mimics the behavior of `WebDriverSelect` but I don't even know if it's really necessary. I propose to remove this for now, and we'll add it back if someone report an issue.

dunglas:
@OndraM all comments fixed.
CS differences was because I followed Symfony ones.

OndraM:
@dunglas Sorry for delays, I will review this soon.

lib/WebDriverCheckbox.php
```diff
@@ -0,0 +1,72 @@
+<?php
```
OndraM: When I was using the classes I was thinking about their names - it seems they are still kinda confusing.

Consider this code:

```php
$checkbox = new WebDriverCheckbox($driver->findElement(...));

$checkbox->deselectAll();
$checkbox->getFirstSelectedOption();
```

It is confusing the class is called "checkbox" (ie. singular), however it actually represents not a single checkbox, but group of checkboxes - and also the methods semantically interacts with multiple objects (deselect*All*, get*First*Selected etc.).

So I have two possible suggestions how to make this more understandable:
1. Name it in plural, like "WebDriverCheckboxes", WebDriverRadios"
2. Or make it obvious it is group of elements: WebDriverCheckboxGroup", "WebDriverRadioGroup"

What do you think?

OndraM:
FYI I was investigating the reason of CI fails, and it seems there is a bug in HtmlUnit browser which is bundled with this version of Selenium server. It seems to be working properly in Firefox & Chrome, but the XPath somehow fails in HtmlUnit. I'll try to update the Selenium server to see if it will be resolved.

dunglas:
Thanks for investigating. For what it worth, tests run locally with the latest version of HtmlUnit.

OndraM:
The CI tests succeeded after Selenium server upgrade, so merging now!

Thanks a lot for the effort @dunglas! (And sorry the review took so long...)

OndraM:
BTW I prepared Wiki page to explain how to use these helper class, would you be interested to add at least some examples for the Checboxes and Radios class (Similarly os it is for the WebDriverSelect)? https://github.com/facebook/php-webdriver/wiki/Select,-checkboxes,-radio-buttons

dunglas:
@OndraM the wiki is readonly, or am I missing something?

OndraM:
It should be editable for contributors - this doesn't work for you? https://github.com/facebook/php-webdriver/wiki/Select%2C-checkboxes%2C-radio-buttons/_edit

dunglas:
It doesn't:

capture d ecran 2018-06-11 a 15 01 39

OndraM:
@dunglas Weird, it used to work for collaborators. Perhaps the wiki permission settings has changed on GitHub, let me check this with @gfosco .

OndraM:
@dunglas The wiki permission should be changed, could you please have a look to see if it works for you?

PR #546, opened 2018-03-14
Allow combined not() and presenceOfElementLocated() functionality

benwhite-deltas:
Currently doing something like the following will cause a PhpUnit test to fail due to an exception thrown by the `presenceOfElementLocated()` call when the element is not present:

```
$driver->wait()->until(
    WebDriverExpectedCondition::not(
        WebDriverExpectedCondition::presenceOfElementLocated(
            WebDriverBy::cssSelector('iframe[name="modal"]')
        )
    )
);
```

This pull request is intended to fix this problem.

Rebased against latest master as previous pull request appears to have failed travis build due to saucelabs timeout.


OndraM:
Thanks @benwhite-deltas!

PR #547, opened 2018-03-14
Complete the PHP CS Fixer config

dunglas:
Some changes requested in your review of #545 can be automatically detected and fixed by PHP CS Fixer.
I updated the config accordingly.

I hope it will save you some time during next reviews :)


OndraM:
:+1: great! A added one more rule and resorted them; will merge after Travis is done.

PR #548, opened 2018-03-16
Fix @see in WebDriverOptions::addCookie

dunglas:
`Cookie` should be either absolute (starting with `\`) or relative to be clickable in PHPStorm.


OndraM:
Thanks :)

PR #549, opened 2018-03-16
Fix the PHPDoc's types of Cookie.php

dunglas:


OndraM:
Cherry-picked, thanks :+1:

PR #560, opened 2018-04-09
W3C-compliant implementation, native geckodriver support

### LINKED ISSUE #469: W3C WebDriver protocol support

dunglas:
This PR provides an implementation of [the W3C flavor of the WebDriver protocol](https://www.w3.org/TR/webdriver/).This implem is good enough to make the [Symfony Panther](https://github.com/symfony/panther)'s test suite green when using the latest version of [geckodriver](https://github.com/mozilla/geckodriver).
It is 100% backward compatible with the current (JsonWire) implementation.

The (only?) missing part is the capabilities support. It's a step to close #469.

TODO:

* [x] Add geckodriver in Travis to run the project's test suite against a valid implementation of the W3C protocol.

Subsequent work (in other PRs?):

* Add new classes for exceptions that wasn't part of JsonWire
* Add support for capabilities and other missing parts

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,10 +34,20 @@ matrix:
       env: DEPENDENCIES="--prefer-lowest"
 
     # Firefox inside Travis environment
-    - php: '7.3'
+    - name: 'Firefox 45 on Travis (OSS protocol)'
+      php: '7.3'
       env: BROWSER_NAME="firefox"
       addons:
         firefox: "45.8.0esr"
+
+    # Firefox with Geckodriver (W3C mode) inside Travis environment
+    - name: 'Firefox latest on Travis (W3C protocol)'
+      php: 7.3
+      env:
+        - BROWSER_NAME="firefox"
+        - GECKODRIVER="1"
+      addons:
+        firefox: latest
 
     # Stable Chrome + Chromedriver inside Travis environment
     - php: '7.3'
--- a/.travis.yml
+++ b/.travis.yml
@@ -108,7 +108,7 @@ install:
 before_script:
   - if [ "$BROWSER_NAME" = "chrome" ]; then mkdir chromedriver; CHROMEDRIVER_VERSION=$(wget -qO- "https://chromedriver.storage.googleapis.com/LATEST_RELEASE"); wget -q -t 3 https://chromedriver.storage.googleapis.com/$CHROMEDRIVER_VERSION/chromedriver_linux64.zip; unzip chromedriver_linux64 -d chromedriver; fi
   - if [ "$BROWSER_NAME" = "chrome" ]; then export CHROMEDRIVER_PATH=$PWD/chromedriver/chromedriver; fi
-  - if [ "$GECKODRIVER" = "1" ]; then mkdir geckodriver; wget -q -t 3 https://github.com/mozilla/geckodriver/releases/download/v0.24.0/geckodriver-v0.24.0-linux64.tar.gz; tar xzf geckodriver-v0.24.0-linux64.tar.gz -C geckodriver; fi
+  - if [ "$GECKODRIVER" = "1" ]; then mkdir geckodriver; wget -q -t 3 https://github.com/mozilla/geckodriver/releases/download/v0.26.0/geckodriver-v0.26.0-linux64.tar.gz; tar xzf geckodriver-v0.26.0-linux64.tar.gz -C geckodriver; fi
   - sh -e /etc/init.d/xvfb start
   - if [ ! -f jar/selenium-server-standalone-3.8.1.jar ]; then wget -q -t 3 -P jar https://selenium-release.storage.googleapis.com/3.8/selenium-server-standalone-3.8.1.jar; fi
   - if [ "$GECKODRIVER" = "1" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -108,9 +108,14 @@ install:
 before_script:
   - if [ "$BROWSER_NAME" = "chrome" ]; then mkdir chromedriver; CHROMEDRIVER_VERSION=$(wget -qO- "https://chromedriver.storage.googleapis.com/LATEST_RELEASE"); wget -q -t 3 https://chromedriver.storage.googleapis.com/$CHROMEDRIVER_VERSION/chromedriver_linux64.zip; unzip chromedriver_linux64 -d chromedriver; fi
   - if [ "$BROWSER_NAME" = "chrome" ]; then export CHROMEDRIVER_PATH=$PWD/chromedriver/chromedriver; fi
+  - if [ "$GECKODRIVER" = "1" ]; then mkdir geckodriver; wget -q -t 3 https://github.com/mozilla/geckodriver/releases/download/v0.24.0/geckodriver-v0.24.0-linux64.tar.gz; tar xzf geckodriver-v0.24.0-linux64.tar.gz -C geckodriver; fi
   - sh -e /etc/init.d/xvfb start
   - if [ ! -f jar/selenium-server-standalone-3.8.1.jar ]; then wget -q -t 3 -P jar https://selenium-release.storage.googleapis.com/3.8/selenium-server-standalone-3.8.1.jar; fi
-  - java -Dwebdriver.firefox.marionette=false -Dwebdriver.chrome.driver="$CHROMEDRIVER_PATH" -jar jar/selenium-server-standalone-3.8.1.jar -enablePassThrough false -log ./logs/selenium.log &
+  - if [ "$GECKODRIVER" = "1" ]; then
+      geckodriver/geckodriver &> ./logs/geckodriver.log &
+    else
+      java -Dwebdriver.firefox.marionette=false -Dwebdriver.chrome.driver="$CHROMEDRIVER_PATH" -jar jar/selenium-server-standalone-3.8.1.jar -enablePassThrough false -log ./logs/selenium.log &
+    fi
   - until $(echo | nc localhost 4444); do sleep 1; echo Waiting for Selenium server on port 4444...; done; echo "Selenium server started"
   - php -S 127.0.0.1:8000 -t tests/functional/web/ &>>./logs/php-server.log &
   - until $(echo | nc localhost 8000); do sleep 1; echo waiting for PHP server on port 8000...; done; echo "PHP server started"
@@ -124,6 +129,7 @@ script:
 after_script:
   - if [ -f ./logs/selenium.log ]; then cat ./logs/selenium.log; fi
   - if [ -f ./logs/php-server.log ]; then cat ./logs/php-server.log; fi
+  - if [ -f ./logs/geckodriver.log ]; then cat ./logs/geckodriver.log; fi
 
 after_success:
   - travis_retry php vendor/bin/php-coveralls -v
--- a/lib/Cookie.php
+++ b/lib/Cookie.php
@@ -190,7 +190,13 @@ class Cookie implements \ArrayAccess
      */
     public function toArray()
     {
-        return $this->cookie;
+        $cookie = $this->cookie;
+        if (!isset($cookie['secure'])) {
+            // Passing a boolean value for the "secure" flag is mandatory when using geckodriver
+            $cookie['secure'] = false;
+        }
+
+        return $cookie;
     }
 
     public function offsetExists($offset)
--- a/lib/Exception/WebDriverException.php
+++ b/lib/Exception/WebDriverException.php
@@ -42,7 +42,7 @@ class WebDriverException extends Exception
     /**
      * Throw WebDriverExceptions based on WebDriver status code.
      *
-     * @param int $status_code
+     * @param int|string $status_code
      * @param string $message
      * @param mixed $results
      *
@@ -85,6 +85,54 @@ class WebDriverException extends Exception
      */
     public static function throwException($status_code, $message, $results)
     {
+        if (is_string($status_code)) {
+            // see https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors
+            switch ($status_code) {
+                case 'no such element':
+                    throw new NoSuchElementException($message, $results);
+                case 'no such frame':
+                    throw new NoSuchFrameException($message, $results);
+                case 'unknown command':
+                    throw new UnknownCommandException($message, $results);
+                case 'stale element reference':
+                    throw new StaleElementReferenceException($message, $results);
+                case 'invalid element state':
+                    throw new InvalidElementStateException($message, $results);
+                case 'unknown error':
+                    throw new UnknownServerException($message, $results);
+                case 'unsupported operation':
+                    throw new ExpectedException($message, $results);
+                case 'element not interactable':
+                    throw new ElementNotSelectableException($message, $results);
+                case 'no such window':
+                    throw new NoSuchDocumentException($message, $results);
+                case 'javascript error':
+                    throw new UnexpectedJavascriptException($message, $results);
+                case 'timeout':
+                    throw new TimeOutException($message, $results);
+                case 'no such window':
+                    throw new NoSuchWindowException($message, $results);
+                case 'invalid cookie domain':
+                    throw new InvalidCookieDomainException($message, $results);
+                case 'unable to set cookie':
+                    throw new UnableToSetCookieException($message, $results);
+                case 'unexpected alert open':
+                    throw new UnexpectedAlertOpenException($message, $results);
+                case 'no such alert':
+                    throw new NoAlertOpenException($message, $results);
+                case 'script timeout':
+                    throw new ScriptTimeoutException($message, $results);
+                case 'invalid selector':
+                    throw new InvalidSelectorException($message, $results);
+                case 'session not created':
+                    throw new SessionNotCreatedException($message, $results);
+                case 'move target out of bounds':
+                    throw new MoveTargetOutOfBoundsException($message, $results);
+                default:
+                    throw new UnrecognizedExceptionException($message, $results);
+            }
+        }
+
         switch ($status_code) {
             case 1:
                 throw new IndexOutOfBoundsException($message, $results);
--- a/lib/Remote/HttpCommandExecutor.php
+++ b/lib/Remote/HttpCommandExecutor.php
@@ -313,8 +313,13 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
 
         $encoded_params = null;
 
-        if ($http_method === 'POST' && $params && is_array($params)) {
-            $encoded_params = json_encode($params);
+        if ($http_method === 'POST') {
+            if ($params && is_array($params)) {
+                $encoded_params = json_encode($params);
+            } elseif ($this->w3cCompliant) {
+                // POST body must be valid JSON in W3C, even if empty: https://www.w3.org/TR/webdriver/#processing-model
+                $encoded_params = '{}';
+            }
         }
 
         curl_setopt($this->curl, CURLOPT_POSTFIELDS, $encoded_params);
--- a/lib/Remote/HttpCommandExecutor.php
+++ b/lib/Remote/HttpCommandExecutor.php
@@ -138,6 +138,29 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
         DriverCommand::TOUCH_UP => ['method' => 'POST', 'url' => '/session/:sessionId/touch/up'],
     ];
     /**
+     * @var array Will be merged with $commands
+     */
+    protected static $w3cCompliantCommands = [
+        DriverCommand::ACCEPT_ALERT => ['method' => 'POST', 'url' => '/session/:sessionId/alert/accept'],
+        DriverCommand::ACTIONS => ['method' => 'POST', 'url' => '/session/:sessionId/actions'],
+        DriverCommand::DISMISS_ALERT => ['method' => 'POST', 'url' => '/session/:sessionId/alert/dismiss'],
+        DriverCommand::EXECUTE_SCRIPT => ['method' => 'POST', 'url' => '/session/:sessionId/execute/sync'],
+        DriverCommand::EXECUTE_ASYNC_SCRIPT => ['method' => 'POST', 'url' => '/session/:sessionId/execute/async'],
+        DriverCommand::GET_CURRENT_WINDOW_HANDLE => ['method' => 'GET', 'url' => '/session/:sessionId/window'],
+        DriverCommand::GET_ELEMENT_LOCATION => ['method' => 'GET', 'url' => '/session/:sessionId/element/:id/rect'],
+        DriverCommand::GET_ELEMENT_PROPERTY => [
+            'method' => 'GET',
+            'url' => '/session/:sessionId/element/:id/property/:name',
+        ],
+        DriverCommand::GET_ELEMENT_SIZE => ['method' => 'GET', 'url' => '/session/:sessionId/element/:id/rect'],
+        DriverCommand::GET_WINDOW_HANDLES => ['method' => 'GET', 'url' => '/session/:sessionId/window/handles'],
+        DriverCommand::GET_ALERT_TEXT => ['method' => 'GET', 'url' => '/session/:sessionId/alert/text'],
+        DriverCommand::IMPLICITLY_WAIT => ['method' => 'POST', 'url' => '/session/:sessionId/timeouts'],
+        DriverCommand::SET_ALERT_VALUE => ['method' => 'POST', 'url' => '/session/:sessionId/alert/text'],
+        DriverCommand::SET_SCRIPT_TIMEOUT => ['method' => 'POST', 'url' => '/session/:sessionId/timeouts'],
+        DriverCommand::SET_TIMEOUT => ['method' => 'POST', 'url' => '/session/:sessionId/timeouts'],
+    ];
+    /**
      * @var string
      */
     protected $url;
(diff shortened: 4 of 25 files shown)


lib/Remote/HttpCommandExecutor.php
```diff
@@ -137,6 +137,13 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
         DriverCommand::TOUCH_SCROLL => ['method' => 'POST', 'url' => '/session/:sessionId/touch/scroll'],
         DriverCommand::TOUCH_UP => ['method' => 'POST', 'url' => '/session/:sessionId/touch/up'],
     ];
+    /**
+     * @var array Wil be merged with $commands
```
mickaelandrieu:
```diff
- Wil
+ Will
```
isn't it?

dunglas: fixed

lib/Remote/JsonWireCompat.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $mechanism = $by->getMechanism();
+        $value = $by->getValue();
+
+        if ($w3cCompliance) {
+            switch ($mechanism) {
+                // Convert to CSS selectors
+                case 'class name':
+                    $mechanism = 'css selector';
+                    $value = ".$value";
```
stof: should it escape special chars (brackets, etc...). I don't know how the JsonWire protocol supports such cases. If it is able to work with them, they need to be escaped when building a CSS selector

stof: the same is true for other cases

dunglas: It's not possible in CSS to escape special chars in identifiers: https://mathiasbynens.be/notes/css-escapes
The only thing we can do is throwing an error.

stof: the article you linked is precisely explaining how to escape special chars to build selectors (see examples in https://mathiasbynens.be/notes/css-escapes#trailing-whitespace for instance). So I don't agree with you

dunglas: Is this valid according to the spec?

> In CSS, identifiers (including element names, classes, and IDs in selectors) can contain only the characters [a-zA-Z0-9] and ISO 10646 characters U+00A0 and higher, plus the hyphen (-) and the underscore (_); they cannot start with a digit, two hyphens, or a hyphen followed by a digit. Identifiers can also contain escaped characters and any ISO 10646 character as a numeric code (see next item). For instance, the identifier "B&W?" may be written as "B\&W\?" or "B\26 W\3F".
https://www.w3.org/TR/CSS21/syndata.html#value-def-identifier

stof: > Identifiers can also contain escaped characters and any ISO 10646 character as a numeric code

When escaping them, you can put what you want

lib/Exception/WebDriverException.php
```diff
@@ -83,8 +84,56 @@ public function getResults()
      * @throws WebDriverCurlException
      * @throws XPathLookupException
      */
-    public static function throwException($status_code, $message, $results)
+    public static function throwException($status_code, $message, $results, $w3c_compliant = false)
```
OndraM: Wouldn't be possible to use polymorphism here and rather have the method implemented as part of something like W3C Compliant Exception?

lib/Remote/HttpCommandExecutor.php
```diff
@@ -242,7 +262,7 @@ public function execute(WebDriverCommand $command)
             }
         }
 
-        if ($params && is_array($params) && $http_method !== 'POST') {
+        if ($params && \is_array($params) && $http_method !== 'POST') {
```
OndraM: I wouldn't change this as part of this PR (also on other lines).

The codestyle may be changed, but then we can do this at once in whole codebase, and do not change it randomly to keep this PR self-contained.

lib/Remote/RemoteWebElement.php
```diff
@@ -325,17 +326,33 @@ public function sendKeys($value)
     {
         $local_file = $this->fileDetector->getLocalFile($value);
         if ($local_file === null) {
-            $params = [
-                'value' => WebDriverKeys::encode($value),
-                ':id' => $this->id,
-            ];
+            if ($this->w3cCompliant) {
```
OndraM: Could this be moved to the compatibility layer as well?

lib/Remote/RemoteWebElement.php
```diff
@@ -325,17 +326,33 @@ public function sendKeys($value)
...
+            }
             $this->executor->execute(DriverCommand::SEND_KEYS_TO_ELEMENT, $params);
         } else {
             $remote_path = $this->upload($local_file);
-            $params = [
-                'value' => WebDriverKeys::encode($remote_path),
-                ':id' => $this->id,
-            ];
+
+            if ($this->w3cCompliant) {
```
OndraM: dtto, maybe put this in compatibility layer instead of the conditional clauses

lib/Remote/RemoteWebElement.php
```diff
@@ -371,6 +388,18 @@ public function setFileDetector(FileDetector $detector)
      */
     public function submit()
     {
+        if ($this->w3cCompliant) {
+            $this->executor->execute(DriverCommand::EXECUTE_SCRIPT, [
```
OndraM: I guess this will face this issue https://github.com/SeleniumHQ/selenium/issues/3398 in some forms:

>  If a form control (such as a submit button) has a name or id of submit it will mask the form's submit method.

See referenced issue for more universal way (using prototype method).

lib/Remote/RemoteWebElement.php
```diff
@@ -371,6 +388,18 @@ public function setFileDetector(FileDetector $detector)
...
+                'script' => sprintf('return arguments[0]%s.submit();', 'form' === $this->getTagName() ? '' : '.form'),
+                'args' => [[JsonWireCompat::WEB_DRIVER_ELEMENT_IDENTIFIER => $this->id]],
+            ]);
+            // Wait for the new page to be ready
+            // TODO: wait for <html> to be ready, or use another more elegant way
+            usleep(10000);
```
OndraM: Why is this here? Isn't possible to `wait()`? This won't be reliable IMHO.

lib/Remote/JsonWireCompat.php
```diff
@@ -0,0 +1,75 @@
...
+        // Legacy JsonWire
+        return $rawElement['ELEMENT'];
+    }
+
+    /**
+     * @param WebDriverBy $by
+     * @param bool $w3cCompliance
+     * @return array
+     */
+    public static function getUsing(WebDriverBy $by, $w3cCompliance)
```
OndraM: `$w3cCompliant` or `$isW3cCompliant` (to have same name as elsewhere)

OndraM:
@dunglas BTW have you seen https://github.com/facebook/php-webdriver/wiki/JsonWireProtocol-vs.-W3C-WebDriver ?

stof:
@OndraM IIRC, the body of the `New session` response allows to identify whether it is a W3C or JsonWire response (and the request for the new session works the same in both protocols IIRC). so this would allow detecting automatically whether the session runs in W3C mode or JsonWire mode.

OndraM:
@stof Right, for example how its done in:
- [ProtocolHandshake](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/ProtocolHandshake.java) in java bindings
- [Handshake and capabilities resolving](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webdriver.py#L252) in python bindings

lib/Cookie.php
```diff
@@ -192,7 +192,13 @@ public function isHttpOnly()
      */
     public function toArray()
     {
-        return $this->cookie;
+        $cookie = $this->cookie;
+        if (null === $cookie['secure']) {
+            // Passing a boolean value for the "secure" flag is mandatory when using geckodriver
```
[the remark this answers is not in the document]
stof: that's indeed a bug. The W3C spec marks it as optional: https://www.w3.org/TR/webdriver1/#dfn-table-for-cookie-conversion

lib/Exception/WebDriverException.php
stof: you don't even need this. Identifying the W3C vs JsonWire status codes is easy, as JsonWire uses numbers, while W3C uses non-numeric strings.

lib/Remote/HttpCommandExecutor.php
dunglas: I'll revert these changes (can't remember why I've done that)

lib/Remote/RemoteWebElement.php
dunglas: Unfortunately I tried to use wait() but IIRC it's too early. I'll take a deeper look.

lib/Remote/HttpCommandExecutor.php
jmalinens: probably phpstorm suggestion :)

lib/Cookie.php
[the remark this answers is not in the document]
dunglas: Yes it's a bug... But we have to workaround it for Geckodriver support.

lib/Remote/RemoteWebElement.php
dunglas: I already use the prototype method. Am I missing something?

lib/Cookie.php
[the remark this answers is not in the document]
stof: have you reported it to the geckodriver team ?

.travis.yml
```diff
@@ -93,8 +93,8 @@ before_script:
   - if [ "$BROWSER_NAME" = "chrome" ]; then mkdir chromedriver; wget -q -t 3 https://chromedriver.storage.googleapis.com/$CHROMEDRIVER_VERSION/chromedriver_linux64.zip; unzip chromedriver_linux64 -d chromedriver; fi
   - if [ "$BROWSER_NAME" = "chrome" ]; then export CHROMEDRIVER_PATH=$PWD/chromedriver/chromedriver; fi
   - sh -e /etc/init.d/xvfb start
-  - if [ ! -f jar/selenium-server-standalone-3.8.1.jar ]; then wget -q -t 3 -P jar https://selenium-release.storage.googleapis.com/3.8/selenium-server-standalone-3.8.1.jar; fi
-  - java -Dwebdriver.firefox.marionette=false -Dwebdriver.chrome.driver="$CHROMEDRIVER_PATH" -jar jar/selenium-server-standalone-3.8.1.jar -enablePassThrough false -log ./logs/selenium.log &
+  - if [ ! -f jar/selenium-server-standalone-3.8.1.jar ]; then wget -q -t 3 -P jar https://selenium-release.storage.googleapis.com/3.14/selenium-server-standalone-3.14.0.jar; fi
```
stof: you need to update the version in the file existence check too.

dunglas:
Protocol handshake added. And tests arre all green locally with Geckodriver.

lib/Cookie.php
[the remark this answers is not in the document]
dunglas: not yet

lib/Remote/RemoteWebElement.php
dunglas: Isn't it better to keep it here. If at some point we remove support for the legacy protocol, it will be easier this way.

.travis.yml
```diff
@@ -92,9 +100,16 @@ install:
...
   - if [ "$BROWSER_NAME" = "chrome" ]; then export CHROMEDRIVER_PATH=$PWD/chromedriver/chromedriver; fi
+  - if [ "$GECKODRIVER" = "1" ]; then mkdir geckodriver; wget -q -t 3 https://github.com/mozilla/geckodriver/releases/download/v0.22.0/geckodriver-v0.22.0-linux64.tar.gz; tar xzf geckodriver-v0.22.0-linux64.tar.gz -C geckodriver; fi
   - sh -e /etc/init.d/xvfb start
-  - if [ ! -f jar/selenium-server-standalone-3.8.1.jar ]; then wget -q -t 3 -P jar https://selenium-release.storage.googleapis.com/3.8/selenium-server-standalone-3.8.1.jar; fi
-  - java -Dwebdriver.firefox.marionette=false -Dwebdriver.chrome.driver="$CHROMEDRIVER_PATH" -jar jar/selenium-server-standalone-3.8.1.jar -enablePassThrough false -log ./logs/selenium.log &
+  - if [ ! -f jar/selenium-server-standalone-3.14.0.jar ]; then wget -q -t 3 -P jar https://selenium-release.storage.googleapis.com/3.14/selenium-server-standalone-3.14.0.jar; fi
+  - |
+    if [ "$GECKODRIVER" = "1" ]
+    then
+      geckodriver/geckodriver &
```
stof: why not running it through the selenium server like others ? You only need to remove the `webdriver.firefox.marionette=false` option you are adding, to go back to the default behavior (using geckdriver for firefox)

lib/Exception/WebDriverException.php
```diff
@@ -42,9 +42,10 @@ public function getResults()
     /**
      * Throw WebDriverExceptions based on WebDriver status code.
      *
-     * @param int $status_code
+     * @param int|string $status_code
      * @param string $message
      * @param mixed $results
+     * @param mixed $w3c_compliant
```
stof: wrong phpdoc. there is no such argument

lib/Remote/HttpCommandExecutor.php
```diff
@@ -317,12 +356,23 @@ public function execute(WebDriverCommand $command)
...
+            $sessionId = $value['sessionId'];
+        } elseif (is_array($results) && array_key_exists('sessionId', $results)) {
+            // Legacy JsonWire
             $sessionId = $results['sessionId'];
         }
 
+        // @see https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors
+        if (isset($value['error'])) {
+            // W3C's WebDriver
+            WebDriverException::throwException($value['error'], $message, $results, true);
```
stof: boolean argument is useless

lib/Remote/JsonWireCompat.php
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        // Legacy JsonWire
+        return $rawElement['ELEMENT'];
+    }
+
+    /**
+     * @param WebDriverBy $by
+     * @param bool $w3cCompliance
+     * @param mixed $w3cCompliant
```
stof: This phpdoc is wrong. We don't have 2 arguments for the W3C compliance (and not a mixed one either)

lib/Remote/RemoteMouse.php
```diff
@@ -27,13 +27,19 @@ class RemoteMouse implements WebDriverMouse
...
      */
     private $executor;
+    /**
+     * @var bool
+     */
+    private $w3cCompliant;
 
     /**
      * @param RemoteExecuteMethod $executor
+     * @param mixed $w3cCompliant
```
stof: `bool`, not `mixed`

lib/Remote/RemoteWebDriver.php
```diff
@@ -110,23 +120,45 @@ public static function create(
...
         if ($required_capabilities !== null) {
             // TODO: Selenium (as of v3.0.1) does accept requiredCapabilities only as a property of desiredCapabilities.
-            // This will probably change in future with the W3C WebDriver spec, but is the only way how to pass these
-            // values now.
+            // This has changed with the W3C WebDriver spec, but is the only way how to pass these
+            // values with the legacy protocol.
             $desired_capabilities->setCapability('requiredCapabilities', $required_capabilities->toArray());
         }
 
+        $parameters = [
```
stof: this should not override the variable entirely, put assign `$parameters['desiredCapabilities']` only

lib/Remote/RemoteWebDriver.php
```diff
@@ -110,23 +120,45 @@ public static function create(
...
 
+        // W3C
+        $parameters = [
+            'capabilities' => [
+                'firstMatch' => [$desired_capabilities->toArray()],
+            ],
+        ];
+
+        // Legacy protocol
+        if (null !== $required_capabilities && $required_capabilities_array = $required_capabilities->toArray()) {
```
stof: is this case really part of the legacy protocol, or should the comment placed above the next `if` instead ?

lib/Remote/RemoteWebDriver.php
```diff
@@ -149,15 +182,17 @@ public static function createBySessionID(
         $connection_timeout_in_ms = null,
         $request_timeout_in_ms = null
     ) {
-        $executor = new HttpCommandExecutor($selenium_server_url);
+        // BC layer to not break the method signature
```
stof: should you trigger a deprecation warning if a child class extends this method without adding the argument ?

lib/Remote/RemoteWebDriver.php
```diff
@@ -519,11 +552,17 @@ public function getCapabilities()
...
      * @param string $selenium_server_url The url of the remote Selenium WebDriver server
      * @param int $timeout_in_ms
+     * @param bool $w3c_compliant
      * @return array
      */
     public static function getAllSessions($selenium_server_url = 'http://localhost:4444/wd/hub', $timeout_in_ms = 30000)
     {
-        $executor = new HttpCommandExecutor($selenium_server_url);
+        // BC layer to not break the method signature
+        if (func_num_args() > 2 && func_get_arg(2)) {
```
stof: Well, forcing people to add a new argument just to get an error looks weird to me.

lib/Remote/RemoteWebElement.php
```diff
@@ -325,20 +348,38 @@ public function sendKeys($value)
     {
         $local_file = $this->fileDetector->getLocalFile($value);
         if ($local_file === null) {
+            if ($this->w3cCompliant) {
+                $params = [
+                    'text' => (string) $value,
+                    ':id' => $this->id,
+                    'foo' => 'bar',
```
stof: do we need this `'foo' => 'bar',` ?

.travis.yml
dunglas: Because it doesn’t behave the same way through Selenium, even with this flag removed.

I’ve no clue why, maybe that Selenium server add some kind of compatibility layer of its own.

lib/Remote/RemoteWebDriver.php
dunglas: I agree, but what do you suggest? Catching the exception and re-throwing another one? Letting the original exception bubbling?

lib/Remote/RemoteWebDriver.php
dunglas: I'm not sure... ping @OndraM

lib/Remote/RemoteWebDriver.php
dunglas: I've not done added it because the library currently doesn't trigger any deprecation. But why not. WDYT @OndraM?

.travis.yml
stof: then, should we have another job running tests through Selenium but with a geckodriver-backed Firefox ? IMO, that's also a common use case (the legacy FirefoxDriver is probably seldom-used today, as it is incompatible with any Firefox version released since months).

lib/Remote/JsonWireCompat.php
dunglas: Escaping is now implemented

dunglas:
Tests are green (including when using Geckodriver directly), and all issues are resolved.
As Geckodriver is working, it's already a large step forward to full W3C compatibility.

Do you think we can merge it as an "experimental" feature @OndraM? We'll still be able to improve things if needed after.

tests/functional/RemoteWebDriverFindElementTest.php
```diff
@@ -61,4 +61,22 @@ public function testShouldFindMultipleElements()
         $this->assertCount(5, $elements);
         $this->assertContainsOnlyInstancesOf(RemoteWebElement::class, $elements);
     }
+
+    public function testEscapeCssSelector()
+    {
+        if (!getenv('GECKODRIVER') === '1') {
+            $this->markTestSkipped('Invalid class identifiers are not supported by the legacy protocol');
```
stof: I would not call it `invalid`. The `class` is not invalid (it is not a valid identifier indeed, and so requires escaping, but the message is not clear IMO).

and if this only applies to classes, maybe the test should be split (to skip only the one about classes)

dunglas: I tried to escape identifiers for the legacy protocol too, but it doesn't work for other reasons. It applies to all identifiers.

wdyt of "CSS selectors containing special characters are not supported by the legacy protocol"?

stof:
There is one failing test here for geckdriver

dunglas:
Tests are green, @OndraM do you think we can merge this one?

OndraM:
@dunglas Wow, nice work! Please give me few days to review the PR as a whole - I hope to get to it later this week.

stof:
@OndraM any news on the review here ?

dunglas:
@OndraM, do you think we can merge this PR?

cpintea:
"selenium grid throw Session [ not available and is not among the last 1000 terminated sessions." when running on MAC/safari nods , Android/real devices nods. 
I notice that this PR is linked with the issue " Session [ not available and is not among the last".
Will this issue be solved by this PR?

lib/Cookie.php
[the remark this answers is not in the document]
dunglas: Chrome now has the same behavior: https://github.com/Codeception/Codeception/commit/ec2fc7362ceea866fb9f5e71181eaa6cfeded8ac#diff-b69c529efcc6368b5c564f8b36665a22R827

We should leave this as is.

lib/Remote/RemoteWebElement.php
dunglas: As it something we want to do (#657), I suggest to leave this as is.

lib/Remote/RemoteWebElement.php
dunglas: Same here, it will make it easier to remove support for the legacy protocol in v2.

dunglas:
Related tests are all green (the failing line in the matrix are unrelated, and are failing for `community` to). Can we merge this and move forward please @OndraM?

dunglas:
@OndraM while I'm here... :) It would be awesome if this patch could be included in the release adding support for Symfony 5 (#670)!

OndraM:
It seems Geckodriver meanwhile started enforcing more parts of W3C protocol - in this case it requires POST body to be always valid JSON (even if empty), as per [Processing model (5.1)](https://www.w3.org/TR/webdriver/#processing-model).

So I updated `HttpCommandExecutor` accordingly - lets see if it fixes the builds :four_leaf_clover: .

dunglas:
And it's now green! Thank you for rebasing and fixing the remaining issues!

OndraM:
Well, finally merged to `community` branch! :tada: I will try to merge few other active PRs now, then tag a new version and get things rolling.

Once again thanks @dunglas for your work!

OndraM:
We also now need to get sorted situation with ChromeOptions capability. The handshake now breaks newer selenium servers - which will not allow `chromeOptions` in `firstMatch` W3C capabilities (because it is against the spec).

```
Facebook\WebDriver\Exception\UnknownServerException: Illegal key values seen in w3c capabilities: [chromeOptions]
```

However if we just rename `chromeOptions` to `goog:chromeOptions`, we would break it for Chromedriver <2.31 which [adds support](https://bugs.chromium.org/p/chromedriver/issues/detail?id=1786) for both of them...

OndraM:
Ok, having some good news. 

Issue with `chromeOptions` was resolved and merged in #673. 

In #674 I also removed workaround which forced Chrome to always use JsonWire OSS protocol. Now we always let the handshake determine the proper protocol. I also added Chromedriver builds to make sure it works with both OSS and W3C protocols - **and the tests are passing** :tada: . (The coverage we have is far from being perfect, so there may be some unresolved issues with W3C protocol - in this case one can explicitly disable W3C protocol on Chromedriver, see issue description in #674.)

So the `community` branch could now officialy undergo some testing of W3C protocol.

PR #574, opened 2018-05-19
Implement web storage (localStorage and sessionStorage)

duncan3dc:
This functionality is available via selenium, and the commands are already defined on `DriverCommand`, so this PR adds classes and methods to make them usable.

I've followed the upstream API as close as possible, the only deviation is how storage is initially accesses, which is via a simple `storage()` method on the `WebDriver` class.
http://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/html5/WebStorage.html

### DIFF
--- a/lib/Html5/LocalStorageInterface.php
+++ b/lib/Html5/LocalStorageInterface.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Facebook\WebDriver\Html5;
+
+/**
+ * Represents the local storage for the site currently opened in the browser.
+ */
+interface LocalStorageInterface
+{
+    /**
+     * Remove all of the items from the storage.
+     */
+    public function clear();
+
+    /**
+     * Retrieve an item from the storage.
+     *
+     * @param string $key
+     *
+     * @return string
+     */
+    public function getItem($key);
+
+    /**
+     * Get all the keys in storage.
+     *
+     * @return string[]
+     */
+    public function keySet();
+
+    /**
+     * Remove a single item from the storage.
+     *
+     * @param string $key
+     */
+    public function removeItem($key);
+
+    /**
+     * Set an item in the storage.
+     *
+     * @param string $key
+     * @param string $value
+     */
+    public function setItem($key, $value);
+
+    /**
+     * Get the number of keys in the storage.
+     *
+     * @return int
+     */
+    public function size();
+}
--- a/lib/Html5/SessionStorageInterface.php
+++ b/lib/Html5/SessionStorageInterface.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Facebook\WebDriver\Html5;
+
+/**
+ * Represents the session storage for the site currently opened in the browser.
+ */
+interface SessionStorageInterface
+{
+    /**
+     * Remove all of the items from the storage.
+     */
+    public function clear();
+
+    /**
+     * Retrieve an item from the storage.
+     *
+     * @param string $key
+     *
+     * @return string
+     */
+    public function getItem($key);
+
+    /**
+     * Get all the keys in storage.
+     *
+     * @return string[]
+     */
+    public function keySet();
+
+    /**
+     * Remove a single item from the storage.
+     *
+     * @param string $key
+     */
+    public function removeItem($key);
+
+    /**
+     * Set an item in the storage.
+     *
+     * @param string $key
+     * @param string $value
+     */
+    public function setItem($key, $value);
+
+    /**
+     * Get the number of keys in the storage.
+     *
+     * @return int
+     */
+    public function size();
+}
--- a/lib/Remote/HttpCommandExecutor.php
+++ b/lib/Remote/HttpCommandExecutor.php
@@ -18,6 +18,11 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
         'Accept: application/json',
     ];
 
+    /** @internal */
+    const SESSION_STORAGE = '/session/:sessionId/session_storage';
+    /** @internal */
+    const LOCAL_STORAGE = '/session/:sessionId/local_storage';
+
     /**
      * @see https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#command-reference
      */
@@ -130,7 +135,25 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
         DriverCommand::TOUCH_MOVE => ['method' => 'POST', 'url' => '/session/:sessionId/touch/move'],
         DriverCommand::TOUCH_SCROLL => ['method' => 'POST', 'url' => '/session/:sessionId/touch/scroll'],
         DriverCommand::TOUCH_UP => ['method' => 'POST', 'url' => '/session/:sessionId/touch/up'],
+
         DriverCommand::CUSTOM_COMMAND => [],
+
+        // Commands which are not part of W3C specification, but remote ends usually implements them
+        DriverCommand::CLEAR_LOCAL_STORAGE => ['method' => 'DELETE', 'url' => self::LOCAL_STORAGE],
+        DriverCommand::CLEAR_SESSION_STORAGE => ['method' => 'DELETE', 'url' => self::SESSION_STORAGE],
+        DriverCommand::GET_LOCAL_STORAGE_ITEM => ['method' => 'GET', 'url' => self::LOCAL_STORAGE . '/key/:key'],
+        DriverCommand::GET_LOCAL_STORAGE_KEYS => ['method' => 'GET', 'url' => self::LOCAL_STORAGE],
+        DriverCommand::GET_LOCAL_STORAGE_SIZE => ['method' => 'GET', 'url' => self::LOCAL_STORAGE . '/size'],
+        DriverCommand::GET_SESSION_STORAGE_ITEM => ['method' => 'GET', 'url' => self::SESSION_STORAGE . '/key/:key'],
+        DriverCommand::GET_SESSION_STORAGE_KEYS => ['method' => 'GET', 'url' => self::SESSION_STORAGE],
+        DriverCommand::GET_SESSION_STORAGE_SIZE => ['method' => 'GET', 'url' => self::SESSION_STORAGE . '/size'],
+        DriverCommand::REMOVE_LOCAL_STORAGE_ITEM => ['method' => 'DELETE', 'url' => self::LOCAL_STORAGE . '/key/:key'],
+        DriverCommand::REMOVE_SESSION_STORAGE_ITEM => [
+            'method' => 'DELETE',
+            'url' => self::SESSION_STORAGE . '/key/:key',
+        ],
+        DriverCommand::SET_LOCAL_STORAGE_ITEM => ['method' => 'POST', 'url' => self::LOCAL_STORAGE],
+        DriverCommand::SET_SESSION_STORAGE_ITEM => ['method' => 'POST', 'url' => self::SESSION_STORAGE],
     ];
     /**
      * @var array Will be merged with $commands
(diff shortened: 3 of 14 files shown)


duncan3dc:
Hi @OndraM, do you have any feedback on this PR? Are there any issues with merging it?

lib/Html5/LocalStorage.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Facebook\WebDriver\Html5;
+
+/**
+ * Represents the local storage for the site currently opened in the browser.
+ */
+interface LocalStorage
+{
+    /**
+     * Remove all of the items from the storage.
```
williamdes, line 11:
```suggestion
     * Remove all of the items from the storage.
     * @return void
```

lib/Html5/SessionStorageInterface.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Facebook\WebDriver\Html5;
+
+/**
+ * Represents the session storage for the site currently opened in the browser.
+ */
+interface SessionStorageInterface
+{
+    /**
+     * Remove all of the items from the storage.
```
williamdes, line 11:
```suggestion
     * Remove all of the items from the storage.
     * @return void
```

williamdes (approved):
:100:

lib/Html5/SessionStorageInterface.php
williamdes: :arrow_up:

lib/Html5/LocalStorageInterface.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Facebook\WebDriver\Html5;
+
+/**
+ * Represents the local storage for the site currently opened in the browser.
+ */
+interface LocalStorageInterface
+{
+    /**
+     * Remove all of the items from the storage.
```
williamdes, line 11:
```suggestion
     * Remove all of the items from the storage.
     * @return void
```

williamdes (approved):
LGTM, 2 missing void return types in phpdoc

OndraM:
> LGTM, 2 missing void return types in phpdoc

Ok, I will have a look.

However I have a different concern - local/session storage is not part of W3C WebDriver. While it surprisingly seems to be working with endpoints from the JsonWire protocol even in W3C mode (at least on browsers we have in CI), for example Java client does it differently - it executes [javascript polyfills](https://github.com/SeleniumHQ/selenium/blob/824617b75181f0cb8a83025d8acfb70b37038f6b/java/client/src/org/openqa/selenium/remote/codec/w3c/W3CHttpCommandCodec.java#L247-L285) when in W3C mode.

I'm not sure whether it is a good idea to add now feature which is not part of W3C (as we aim to make php-webdriver compatible with WebDriver).

OndraM:
:top: Whats is you opinion, @williamdes @andrewnicols @dunglas @driesvints? Do you think we should include this in php-webdriver, even though it is not part of W3C WebDriver (but it is implemented in Java bindings)? 

Or this should be solved by users using javascript?

andrewnicols:
While the patch itself looks okay (a few things to address), I do agree with you Ondřej that this is something that is probably not right for php-webdriver.

I'm sure this will not be the only example of commands which some drivers have chosen to implement outside of the specification, but if we implement this now then we have a couple of issues to solve in future:
1. if this is included in a version of the specification but with a slightly different API then we will have to manage that situation;
2. if we implement this one, then why not the next optional command set... or the one after that? And what if some of those are browser-specific, or if different browsers implement them differently? Accepting this sets a precedent.

If this is to land then I'm not sure about the use of the constant for the URL prefix on the new DriverCommands. We don't do this elsewhere as far as I'm aware. Also, I suspect that the URLs in the local storage commands are wrong because they're all at `/session/:sessionId/session_storage` when I would expect to see `local_storage` as the endpoint.

Rather than accepting this patch, perhaps we should create a way to add optional extensions to php-webdriver for cases like this where a widely-implemented feature exists but is not a part of the specification. It should be relatively easy to write this.

At the moment the main change to the core WebDriver code is to add the new Driver command configuration into the `HttpCommandExecutor`, and a new `storage()` function:

```
public function storage()
{
    return new RemoteWebStorage($this->getExecuteMethod());
}
```

Every other change is handled from that `RemoteWebStorage` class via the Local and Remote storage handlers.

We could introduce a new Interface like `WebDriverExtension` with a constructor and method to fetch the command configuration, and then add to `RemoteWebDriver` with something like:
```
// lib/WebDriverExtension.php
interface WebDriverExtension
{
    public function __construct(RemoteExecuteMethod $executor, WebDriver $driver, $isW3cCompliant = false);
    public static function getCommands(): array;
}

// lib/Remote/RemoteWebDriver.php
public function extension(string $className)
{
    if (!array_key_exists($extension, $this->extensions)) {
        if (!is_a($className, WebDriverExtension::class, true)) {
            throw new \InvalidArgumentException('...');
        }
        $this->extensions[$className] = new $className($this->getExecuteMethod(), $this, $this->isW3cCompliant);
        $this->executor->registerExtensionCommands($className::getCommands());
    }

    return $this->extensions[$className];
}
```

We would also need a corresponding way to register the driver commands in `HttpCommandExecutor`:
```
// lib/Remote/HttpCommandExecutor.php
public static function registerExtensionCommands(array $extensionCommands): void
{
    self::$commands = array_merge(self::$commands, $extensionCommands);
}
```

The extension would look like:
```
// lib/Testwork/WebDriverExtensions/Storage.php
<?php
namespace Acme\Testwork\WebdriverExtensions;

use Acme\Testwork\WebdriverExtensions\Storage\LocalSessionStorage;
use Acme\Testwork\WebdriverExtensions\Storage\RemoteSessionStorage;
use Facebook\WebDriver\WebDriverExtension;

class Storage implements WebDriverExtension
{
    public function __construct(RemoteExecuteMethod $executor, WebDriver $driver, $isW3cCompliant = false)
    {
        $this->executor = $executor;
        $this->driver = $driver;
        $this->isW3cCompliant = $isW3cCompliant;
    }

    public static function getCommands(): array
    {
        return [
            RemoteSessionStorage::CLEAR_SESSION_STORAGE => ['method' => 'DELETE', 'url' => '/session/:sessionId/session_storage'],
            RemoteSessionStorage::GET_SESSION_STORAGE_ITEM => ['method' => 'GET', 'url' => '/session/:sessionId/session_storage/key/:key'],
            RemoteSessionStorage::GET_SESSION_STORAGE_KEYS => ['method' => 'GET', 'url' => '/session/:sessionId/session_storage'],
            RemoteSessionStorage::GET_SESSION_STORAGE_SIZE => ['method' => 'GET', 'url' => '/session/:sessionId/session_storage/size'],
            RemoteSessionStorage::REMOVE_SESSION_STORAGE_ITEM => [
                'method' => 'DELETE',
                'url' => '/session/:sessionId/local_storage/key/:key',
            ],
            RemoteSessionStorage::SET_SESSION_STORAGE_ITEM => ['method' => 'POST', 'url' => '/session/:sessionId/session_storage'],
            RemoteLocalStorage::CLEAR_LOCAL_STORAGE => ['method' => 'DELETE', 'url' => '/session/:sessionId/local_storage'],
            RemoteLocalStorage::GET_LOCAL_STORAGE_ITEM => ['method' => 'GET', 'url' => '/session/:sessionId/local_storage/key/:key'],
            RemoteLocalStorage::GET_LOCAL_STORAGE_KEYS => ['method' => 'GET', 'url' => '/session/:sessionId/local_storage'],
            RemoteLocalStorage::GET_LOCAL_STORAGE_SIZE => ['method' => 'GET', 'url' => '/session/:sessionId/local_storage/size'],
            RemoteLocalStorage::REMOVE_LOCAL_STORAGE_ITEM => ['method' => 'DELETE', 'url' => '/session/:sessionId/local_storage/key/:key'],
            RemoteLocalStorage::SET_LOCAL_STORAGE_ITEM => ['method' => 'POST', 'url' => '/session/:sessionId/local_storage'],
        ];

        // This could also be done as:
        return array_merge(
            LocalSessionStorage::getCommands(),
            RemoteLocalStorage::getCommands()
        );
    }

    public function getLocalStorage()
    {
        return new RemoteLocalStorage($this->executor);
    }

    public function getSessionStorage()
    {
        return new RemoteSessionStorage($this->executor);
    }
}

```

From the calling code it could then be called:
```
<?php

namespace Acme\Testwork;

use Acme\Testwork\WebdriverExtensions\Storage;

// ...
$driver = RemoteWebDriver::create($host, DesiredCapabilities::chrome());

// ...
$driver->extension(Storage::class)->getLocalStorage()->clear();
$driver->extension(Storage::class)->getLocalStorage()->get();
```

lib/Remote/HttpCommandExecutor.php
```diff
@@ -18,6 +18,11 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
         'Accept: application/json',
     ];
 
+    /** @internal */
+    const SESSION_STORAGE = '/session/:sessionId/session_storage';
+    /** @internal */
+    const LOCAL_STORAGE = '/session/:sessionId/session_storage';
```
andrewnicols, line 24: This URL looks wrong to me.

Also, not sure that we should use const here at all.

OndraM: This was my edit of the original PR, and I made the mistake :facepalm: .

Const made sense to me, as this part is repeated for many times here.

andrewnicols: heh fair enough - it just drew my eye because it is the first time we've used a constant as part of the URL in the command.

driesvints:
@OndraM I'm not too familiar with this so not sure I'm of much help here, sorry.

williamdes:
>  Whats is you opinion, @williamdes @andrewnicols @dunglas @driesvints? Do you think we should include this in php-webdriver, even though it is not part of W3C WebDriver (but it is implemented in Java bindings)?
> 
> Or should leave this on users (to implement it using javascript)?

I agree it should be implemented, the URLs exist on the API.
Users **sometimes** write bad and broken code to manage things we do not implement.

williamdes (approved):

OndraM:
Ok, so, as WebStorage endpoints are not part of W3C WebDriver, and eg. chromedriver actively **rejects** request on non-W3C endpoints when in W3C mode, I'm closing this PR, as the implementation will not work.

It is easy to workaround this using `executeScript()`:

```php
$localStorageValue = $driver->executeScript(sprintf("return localStorage.getItem('%s')", $key));
// ... etc.
```

It would be also possible to implement local/session storage wrapper implementing Countable and ArrayAccess interfaces for convenience. I tried for example something like this:
```php
<?php

use Facebook\WebDriver\Html5\LocalStorageInterface;
use Facebook\WebDriver\JavaScriptExecutor;

/**
 * Executes the commands to access HTML5 localStorage on the remote webdriver server.
 */
class LocalStorage implements \Countable, \ArrayAccess
{
    /**
     * @var JavaScriptExecutor
     */
    private $executor;

    /**
     * @param JavaScriptExecutor $executor
     */
    public function __construct(JavaScriptExecutor $executor)
    {
        $this->executor = $executor;
    }

    public function clear()
    {
        $this->executor->executeScript('localStorage.clear()');
    }

    public function offsetGet($key)
    {
        return $this->executor->executeScript(sprintf("return localStorage.getItem('%s')", $key));
    }

    public function offsetExists($key)
    {
        return $this->offsetGet($key) !== null;
    }

    public function allKeys()
    {
        return $this->executor->executeScript('return Object.keys(localStorage)');
    }

    public function offsetUnset($key)
    {
        return $this->executor->executeScript(sprintf("localStorage.removeItem('%s')", $key));
    }

    public function offsetSet($key, $value)
    {
        return $this->executor->executeScript(sprintf("localStorage.setItem('%s', '%s')", $key, $value));
    }

    public function count()
    {
        return $this->executor->executeScript('return localStorage.length');
    }
}
```
```php
$localStorage = new LocalStorage($driver);

// store data
$localStorage['my-key'] = 'my data';
// access data
$data = $localStorage['my-key'];
// unset key
unset($localStorage['my-key']);

// etc.
```

While this hypothetically be part of php-webdriver, I don't think its our job to implement such features (php-webdriver is mainly PHP language bindings for W3C protocol). But this could rather be part of some high-level library built on top of php-webdriver (Codeception, Dusk etc.)

PR #575, opened 2018-05-25
Mozilla firefox geckodriver fix.

zyrukas:
I was unable to make clicks in selenium + geckodriver, until I made this change.
Mozilla firefox geckodriver fix. ([PHPUnit\Framework\Exception] Undefined index: ELEMENT).

### DIFF
--- a/lib/Remote/RemoteWebDriver.php
+++ b/lib/Remote/RemoteWebDriver.php
@@ -187,7 +187,7 @@ class RemoteWebDriver implements WebDriver, JavaScriptExecutor, WebDriverHasInpu
             $params
         );
 
-        return $this->newElement($raw_element['ELEMENT']);
+        return $this->newElement($raw_element['ELEMENT'] ?? $raw_element[\key($raw_element)]);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


zyrukas:
@OndraM Please review :)

lib/Remote/RemoteWebDriver.php
```diff
@@ -187,7 +187,9 @@ public function findElement(WebDriverBy $by)
             $params
         );
 
-        return $this->newElement($raw_element['ELEMENT']);
+        return $this->newElement(
+            isset($raw_element['ELEMENT']) ? $raw_element['ELEMENT'] : $raw_element[\key($raw_element)]
```
wolzogen: Hi, you can create a separate function. In order not to violate DRY

```php
private function _getElement($raw_element) {
    return isset($raw_element['ELEMENT']) ? 
        $raw_element['ELEMENT'] : $raw_element[\key($raw_element)];
}
```

lib/Remote/RemoteWebDriver.php
```diff
@@ -207,7 +209,9 @@ public function findElements(WebDriverBy $by)
 
         $elements = [];
         foreach ($raw_elements as $raw_element) {
-            $elements[] = $this->newElement($raw_element['ELEMENT']);
+            $elements[] = $this->newElement(
```
wolzogen:
```php
$elements[] = $this->newElement($this->_getElement($raw_element));
```

OndraM:
Hi @zyrukas , thanks for this, but this will fix only one particular issue - but there is much more that needs to be done in order to get everything working in latest FF & geckodriver and W3C protocol. See #469. I guess this will be replaced by #560 , am I right @dunglas ?

WinstonN:
Hi Hi, 

For what it's worth, I got this error using Codeception & BrowserStack using FF v62.0
Using FF v58.0 I no longer experience the error

PR #577, opened 2018-05-30
Check for BC breaks during Travis build

OndraM:


OndraM:
This cannot be done now, because of conflicts in dependencies (cause originally by use of PHPUnit 5.x). But should be revisited later.

PR #588, opened 2018-07-15
Create index.php to include vendor/autoload.php

GregJohnStewart:
Needed for easy usable phar creation. Default phar will point to 'index.php', instead of what actually loads the correct files ('vendor/autoload.php').


OndraM:
Closing, as discussed in https://github.com/facebook/php-webdriver/issues/586#issuecomment-405418429

PR #591, opened 2018-07-30
added PHPDoc to RemoteWebDriver's execute method

itaibh:
In order to have `@throws WebDriverException`, so IDEs don't complain about unused `catch` statements.

### DIFF
--- a/lib/Remote/RemoteWebDriver.php
+++ b/lib/Remote/RemoteWebDriver.php
@@ -535,6 +535,12 @@ class RemoteWebDriver implements WebDriver, JavaScriptExecutor, WebDriverHasInpu
         return $executor->execute($command)->getValue();
     }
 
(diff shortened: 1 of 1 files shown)


OndraM:
We are generally not using `@throws` everywhere... Adding it here would mean IDEs will start to complain eveywhere else where `execute` is used internally (`findElement()`, `get()`, `close()` and tons of other methods) :thinking: .

I'm not sure how to handle this. Best would maybe for all user-space methods to have defined `@throws WebDriverException`. But I don't know whether we want to update this in 1.x. 

What do you think @itaibh ?

OndraM:
Closing this for now - however we should take care of this in php-webdriver 2.0 (see #657), as this is simply too much effort doing this properly for 1.x.

PR #599, opened 2018-09-18
W3c

mpierzchalski:
added W3C support

tested for: 
* selenium-3.14.0
* geckodriver-0.21
* latest Firefox
* latest Chrome


mpierzchalski:
@williamdes it is hard to do it. I just translate all **JsonWireProtocol** commands into **W3C**. In that meaning all commands are implemented. But I do not give you 100% sure, that all of them work properly, because functional tests covers only bigger part of them :/ I can try to describe the translation by comparing both protocols, maybe this way would be more precise. 

The main goal for me to add this W3C support, was to do not change interface of WebDriver, to keep all that changes hidden in source. So I do not use different commands depends on protocol, for ex. in case when user wants to change window's size, he do it in same way like before.
```
$driver->manage()->window()->setSize(new WebDriverDimension(100, 200));
```
The WebDriver after session being create, guesses in what dialect receive the response, and keep that information in memory. In case of W3C, the WebDriver will translate every single next command before its execution into W3C.

OndraM:
Hi @mpierzchalski , thanks for you PR :+1: !
Have you seen PR #560 , where W3C partial support is already being implemented? I suggest merging the efforts - could you please have look and see if @dunglas won't seek any help with the PR?

Thanks :100:

dunglas:
@mpierzchalski don't hesitate to ping me on the Symfony Slack (or on Twitter)!

dunglas:
@OndraM, this one and #560 take a very different path:

- this one use an abstract protocol that is then translated in the W2C or the JSONWire version
- mine directly send the query in the good flavor regarding which protocol should be used

Also, it looks like this one is fully compatible with the current PHP API (tests and method signatures are changed), while mine tried to preserve BC.

Consequently, it's not possible to merge both PR, the approach is too different. Which one do you prefer we focus on?

jmalinens:
nice work @mpierzchalski  and @dunglas!

we are eager to use it when merged!

OndraM:
This was in the end superseded by #560 . However big kudos to @mpierzchalski for working on  this! :1st_place_medal:

PR #602, opened 2018-10-01
Use polyfill-mbstring when ext-mbstring isn't available

dunglas:
So, more users will be able to use the project.


dunglas:
Errors are unrelated. This can be merge (/cc @OndraM)

OndraM:
Rebased and build passed, merging now. Thanks @dunglas!

PR #606, opened 2018-10-18

kloentje2:
For debugging purposes most of the users I know are using .local domains without a valid SSL certificate. This will accept those certificates


OndraM:
.local domain is [reserved for Multicast](https://en.wikipedia.org/wiki/.local).

You should change your development domain to `.localhost`, which will always translate to 127.0.0.1 and does not require https - see https://tools.ietf.org/html/rfc6761#section-6.3

Issue #605, opened 2018-10-15
Allow executing of custom webdriver commands


matej-pavla:
### What are you trying to achieve? (Expected behavior)

To execute chromedriver's  `Page.setDownloadBehavior` to allow downloading of files when using chrome in `--headless` mode. More generally to allow executing of custom webdriver commands.

More about the chromedriver issue here: https://bugs.chromium.org/p/chromedriver/issues/detail?id=2454 Basically, to be able to download files in headless chrome, one needs to execute command to url `/session/:sessionId/chromium/send_command` with following params:

```php
$params = array(
        'cmd' => 'Page.setDownloadBehavior',
        'params' => array(
                "behavior" => 'allow',
                "downloadPath" => DOWNLOADS_PATH,
        ),
);
```

Using existing webdriver's functionality it's not possible, since `HttpCommandExecutor`'s `execute` which seem to contain the logic for sending the request only works with the pre-defined commands from the `$commands` array and `/session/:sessionId/chromium/send_command` is just not one of them.

To solve this problem it would be nice to add some method or a way to easily extend `HttpCommandExecutor` to handle custom commands and then be able to use it with `RemoteWebDriver`.

### What do you get instead? (Actual behavior)
I am not able to currently execute such command via php-webdriver.

### Details

* Php-webdriver version: 1.6
* PHP version: 7.1
* Selenium server version: 3.14
* Operating system: Mac
* Browser used + version: Chrome

era404design:
Sorry, was there a resolution for this? I haven't been able to perform a successful download headlessly, whereas all downloads performed non-headlessly are successful. Thanks in advance.

OndraM:
Custom commands were merged in #738 and will be part of 1.8.0 - so it may already by possible to achieve this via custom command.
Furthermore there is also PR #753 for DevTools commands which will make this even more convenient in future.

PR #611, opened 2018-10-29
Support of custom http WebDriver commands

maxvgi:
This patch allows to send user-defined http-requests.

Can be used to send browser-specific commands or to work with selenium-compatible server, e.x. Appium.

Expects to solve #605.

Code example:

```
clipboard = 
  $driver->executeCustomCommand(
    '/session/:sessionId/appium/device/get_clipboard/',
    'POST'
  );
```


OndraM:
Nice idea, @maxvgi!

@dunglas you know the library internals well, what is you opinion? Is this a good way how to provide a way to send custom commands (to eg. some proprietary vendor specific endpoint)? Any suggestions?

maxvgi:
> Hi @maxvgi , I made same internal changes and refactored this PR to #738. What do you think?

Thank you!
Well done!

PR #618, opened 2018-12-11
Correct return type

phamuyentri:
Because return type of interface `ExecuteMethod::execute()` is `WebDriverResponse`, so return type of these methods should be `WebDriverResponse` too.


OndraM:
Thanks for the PR, @phamuyentri!

However you came across on of the issues in current codebase - the typehints are not accurate. Actually both of the methods `getLog`  and `getAvailableLogTypes` does return an array. We hope to get this inconsistencies fixed in php-webdriver 2.0 - see #657

PR #621, opened 2018-12-31
improved xpath that matches related elements

oleg-andreyev:
It's possible to put <input/> into a form, but Element will be associated with a different form

Ref.: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#form

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ This project versioning adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - `WebDriverCheckboxes` and `WebDriverRadios` helper classes to simplify interaction with checkboxes and radio buttons.
 
+### Fixed
+- Stop sending null values in Cookie object, which is against the protocol and may cause request to remote ends to fail.
+
+### Changed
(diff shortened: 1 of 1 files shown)


oleg-andreyev:
@gfosco @OndraM could you please take a look, thanks in advance

OndraM:
These classes were recently contributed by @dunglas - maybe he can have have a look as well, please?

OndraM:
Rebased against current community branch; will be merged after CI build finished.

OndraM:
Well, according to Coveralls, these lines are now uncovered by tests:
https://coveralls.io/builds/23884998/source?filename=lib%2FAbstractWebDriverCheckboxOrRadio.php#L215

But they were covered in latest community branch build: https://coveralls.io/builds/23884443/source?filename=lib/AbstractWebDriverCheckboxOrRadio.php

Is this now a dead code, bug in the implementation or false positive in the coverage report?

OndraM:
Ok, it seems unrelated to this pull-request. However I extended the testcases to cover this as well.

OndraM:
Merged to community branch manually, thanks!

PR #627, opened 2019-02-25
626 no null values in cookie

JorisVanEijden:
Both JsonWireProtocol and w3c protocol say to leave an entry off
rather than having a null value.

https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol
https://w3c.github.io/webdriver/#add-cookie

### DIFF
--- a/lib/Cookie.php
+++ b/lib/Cookie.php
@@ -27,15 +27,7 @@ use InvalidArgumentException;
 class Cookie implements \ArrayAccess
 {
     /** @var array */
-    protected $cookie = [
-        'name' => null,
-        'value' => null,
-        'path' => null,
-        'domain' => null,
-        'expiry' => null,
-        'secure' => null,
-        'httpOnly' => null,
-    ];
+    protected $cookie = [];
 
     /**
      * @param string $name The name of the cookie; may not be null or an empty string.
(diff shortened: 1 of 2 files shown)


JorisVanEijden:
The travis fail seems to be unrelated to this PR:
```bash
1) Facebook\WebDriver\WebDriverTimeoutsTest::testShouldFailGettingDelayedElementWithoutWait
Failed asserting that exception of type "Facebook\WebDriver\Exception\NoSuchElementException" is 
thrown.
```
How does this happen? And what can I do to fix it?

The code coverage decreased even though every single line I touched is tested and I added an extra test. How does this happen? And what can I do to fix it?

derekmd:
With Chromedriver 74 binaries released, it now causes some cookie requests to fail without this pull request being merged and tagged.

https://chromium-review.googlesource.com/c/chromium/src/+/1366925/3/chrome/test/chromedriver/window_commands.cc#1564

As a workaround, you must manually set every cookie value to avoid `null` Set-Cookie items being sent.

OndraM:
Thanks @JorisVanEijden for the catch. I rebased the branch against current community branch, let's see if the Travis build now pass. If so, I will get it merged.

OndraM:
Merged, thanks @JorisVanEijden! :100:

PR #643, opened 2019-06-11
Fix regression in 1.7.0 - call to a member function toArray() on array

### LINKED ISSUE #642: [Error] Call to a member function toArray() on array
Working tests with Codeception 2.5.6 + WebDriver module in pair with npm webdriver-manager + facebook/webdriver inside a Chrome Browser

After update composer dependencies with `facebook/webdriver 1.6 > 1.7` - I encounter an error on test\suite start even before browser is up.
                                                        
```
  [Error] Call to a member function toArray() on array  
                                                        
#1  /var/lib/jenkins/workspace/project/testing/vendor/facebook/webdriver/lib/Remote/RemoteWebDriver.php:112
```

Configure: 
Acceptance yaml configuration to run Chrome is 👍 
```
    - WebDriver:
        url: 'https://localhost/'
        wait: 1
        browser: chrome
        window_size: false
        capabilities:
          chromeOptions:
            args: ["--disable-gpu", "--window-size=1920,2500"] #"--headless",

```
Test example is Cest format:
```
public function testMe(AcceptanceTester $I) {
$I->amOnPage('/login');
}
```

### Details

```
composer.json:
require: {
        "facebook/webdriver": "1.6",
        "codeception/codeception": "^2.4",
}
```
* Php-webdriver version: 1.7
* PHP version: 7.1
* Selenium server version: selenium-server-standalone-3.141.59.jar
* Operating system: Windows 10 (1809)
* Browser used + version: Chrome Version 75.0.3770.80 (Official Build) (64-bit)  chromedriver_75.0.3770.8

Lctrs:
Close #642 

DesiredCapabilities#toArray() is a mutable operation which can lead to undesirable side effect. Cloning the instance before calling it will prevent this.

Exemple of side effect : 
```php
<?php

use Facebook\WebDriver\Chrome\ChromeOptions;
use Facebook\WebDriver\Remote\DesiredCapabilities;

require __DIR__.'/vendor/autoload.php';

$capabilities = new DesiredCapabilities();
$capabilities->setCapability(ChromeOptions::CAPABILITY, new ChromeOptions());
$options = $capabilities->getCapability(ChromeOptions::CAPABILITY); // Instance of Facebook\WebDriver\Chrome\ChromeOptions
$capabilities->toArray();
$options = $capabilities->getCapability(ChromeOptions::CAPABILITY); // Oh no it's an array
```

### DIFF
--- a/lib/Remote/DesiredCapabilities.php
+++ b/lib/Remote/DesiredCapabilities.php
@@ -156,6 +156,7 @@ class DesiredCapabilities implements WebDriverCapabilities
     }
 
     /**
+     * @todo Remove side-effects - not change ie. ChromeOptions::CAPABILITY from instance of ChromeOptions to an array
      * @return array
      */
     public function toArray()
--- a/lib/Remote/RemoteWebDriver.php
+++ b/lib/Remote/RemoteWebDriver.php
@@ -109,8 +109,10 @@ class RemoteWebDriver implements WebDriver, JavaScriptExecutor, WebDriverHasInpu
             $currentChromeOptions = $desired_capabilities->getCapability(ChromeOptions::CAPABILITY);
             $chromeOptions = !empty($currentChromeOptions) ? $currentChromeOptions : new ChromeOptions();
 
-            if (!isset($chromeOptions->toArray()['w3c'])) {
+            if ($chromeOptions instanceof ChromeOptions && !isset($chromeOptions->toArray()['w3c'])) {
                 $chromeOptions->setExperimentalOption('w3c', false);
+            } elseif (is_array($chromeOptions) && !isset($chromeOptions['w3c'])) {
+                $chromeOptions['w3c'] = false;
             }
 
             $desired_capabilities->setCapability(ChromeOptions::CAPABILITY, $chromeOptions);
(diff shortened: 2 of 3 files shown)


OndraM:
True, `toArray()` should not work this way... This is unexpected and is something which should be taken care of in next major version.

I amended some TODO notices in the code so that we don't forget on this.

Thanks, I will merge it and tag 1.7.1 after CI builds are done.

Lctrs:
@OndraM We can already turn `toArray()` into an immutable operation easily : 

```php
    /**
     * @return array
     */
    public function toArray()
    {
        $capabilities = $this->capabilities;

        if (isset($capabilities[ChromeOptions::CAPABILITY]) &&
            $capabilities[ChromeOptions::CAPABILITY] instanceof ChromeOptions
        ) {
            $capabilities[ChromeOptions::CAPABILITY] =
                $capabilities[ChromeOptions::CAPABILITY]->toArray();
        }

        if (isset($capabilities[FirefoxDriver::PROFILE]) &&
            $capabilities[FirefoxDriver::PROFILE] instanceof FirefoxProfile
        ) {
            $capabilities[FirefoxDriver::PROFILE] =
                $capabilities[FirefoxDriver::PROFILE]->encode();
        }

        return $capabilities;
    }
```

I don't know if that can be considered as a BC.

OndraM:
@Lctrs Right, but I'm afraid of the BC - the library is not generally type-safe and someone could rely on this mutable behavior. 

This is also reason I in the end removed the clone and just check if ChromeOptions is already an array or not, and act accordingly. This is not the best from code point of view, but it should handle both of the cases that may now occur and does not introduce any change in previous behavior.

boboldehampsink:
When will this be released - it's broken for me now

OndraM:
@boboldehampsink Quite soon, I'm on in right now. But in the meantime you can use dev-community depencency.

PR #648, opened 2019-06-26
add missing json wire protocol for status command

WhatCMSAdmin:
'status' is defined in the DriverCommand class but it's not usable unless the wire protocol is defined in HttpCommandExecutor.


OndraM:
Replaced by #689 - thanks @WhatCMSAdmin for contribution!

PR #652, opened 2019-07-19
updating to latest chromeWebdriver options structure

abdrmdn:
Chrome WebDriver has changed the chrome options lately which leaves this repo a broken dependency to all other modules using this as a base.
`CAPABILITY` should 'goog:chromeOptions' instead of 'chromeOptions'

### DIFF
--- a/lib/Chrome/ChromeOptions.php
+++ b/lib/Chrome/ChromeOptions.php
@@ -27,7 +27,7 @@ class ChromeOptions
     /**
      * The key of chrome options in desired capabilities.
      */
-    const CAPABILITY = 'chromeOptions';
+    const CAPABILITY = 'goog:chromeOptions';
     /**
      * @var array
      */


abdrmdn:
Related issue in here
https://github.com/facebook/php-webdriver/issues/653

attilix:
Great! We have a fix, when wil this be merged?

abdrmdn:
@OndraM  How do we get this merged?

OndraM:
Hi @abdrmdn, thanks for the contribution!

However Chromedriver still supports `chromeOptions`. And simply changing it to `goog:chromeOptions` will break support for Chromedrivers older than [version 2.31, where the support was added](https://bugs.chromium.org/p/chromedriver/issues/detail?id=1786) ... There must be some kind of workaround (which I'm working on right now] to support both of the keys, depending on protocol.

abdrmdn:
@OndraM  Great, thanks man!

PR #663, opened 2019-09-19
Makes it work with Chrome 77

wagnerjsilva:
Error caused by Notice: Undefined index: ELEMENT


OndraM:
Latest tagged version of php-webdriver (1.7.1) does not use W3C protocol, so upgrading to this version should already fix it - see #640 .

Also this will be fixed by next release of php-webdriver even for W3C protocol - see #560 .

PR #669, opened 2019-10-25
Bugfix. Undefined index: ELEMENT, from Codeception

vommuan:
https://github.com/facebook/php-webdriver/issues/469


OndraM:
Hi @vommuan, thanks fort he effort. However we do have this already implemented as part of #560 - with many other parts of W3C protocol support.

PR #670, opened 2019-10-28
Compatibility with Symfony 5

dunglas:
Symfony 5 is about to be released, this PR allows to use PHP WebDriver (and so Panther) with it. There is no code changes, the new version only has to be allowed in `composer.json`.

@OndraM could you tag a new release after merging please 🙏?


OndraM:
Thanks @dunglas, I will tag this together with #560 👌

dunglas:
Awesome! Thanks

Issue #573, opened 2018-05-17
useragent not apply on Chrome with docker Selenium


nono0481:
### What are you trying to achieve? (Expected behavior)
I want to set a UserAgent and MobileEmulation when i start a browser on selenium.
For this, i'm using ChromeOption class.

In fact, Google Chrome Driver has changed the option name to manipulate browser arguments. Since version 2.31, the option name is not chromeOptions but goog:chromeOptions : 
* https://github.com/elgalu/docker-selenium/issues/201
* https://chromedriver.storage.googleapis.com/2.31/notes.txt

When I change the name in ChromeOptions class, it's works fine.
=> Can you update the code please ?

### What do you get instead? (Actual behavior)
The user-agent is set on chromeOptions but when the browser is stated the user-agent is not applied

### How could the issue be reproduced? (Steps to reproduce)
* Start a Selenium Grid hub with docker : https://hub.docker.com/r/selenium/hub/
* Start a Selenium Node Chrome with docker : https://hub.docker.com/r/selenium/node-chrome-debug/
* Run a test with a custom User-Agent :

```php
$desired_capabilities = DesiredCapabilities::chrome();
$options = new ChromeOptions();
$options->addArguments(array('--user-agent=Mozilla/5.0(iPad; U; CPU iPhone OS 3_2 like Mac OS X Tests_Autos; en-us) AppleWebKit/531.21.10 (KHTML, like Gecko) Version/4.0.4 Mobile/7B314 Safari/531.21.10'));
$desired_capabilities->setCapability(ChromeOptions::CAPABILITY, $options);
$this->webDriver = RemoteWebDriver::create($selenium_server, $desired_capabilities);
```

### Details
* Php-webdriver version: 1.5
* PHP version: 5.6
* Selenium server version: 3.11
* Operating system: Ubuntu 16.04
* Browser used + version: Google Chrome 66.0.3359.181

crazyfree:
I also can't set mobileEmulation 
```php

$options = (new ChromeOptions)->addArguments([
            '--disable-gpu',
            '--headless'
        ]);

$mobileEmulation = ['deviceName' => 'Nexus 5'];
$options->setExperimentalOption('mobileEmulation', $mobileEmulation);

return RemoteWebDriver::create( // selenium-hub
            'http://selenium-hub:4444/wd/hub',
            $desiredCapabilities
        );
```

Kindly support

OndraM:
This should be resolved by #673 and fixed in next php-webdriver release.

Issue #551, opened 2018-03-20
Compatibility with ChromeDriver 2.31


freiondrej:
### What are you trying to achieve? (Expected behavior)

setting chromeOptions

### What do you get instead? (Actual behavior)

the options are ignored

### How could the issue be reproduced? (Steps to reproduce)

With ChromeDriver 2.31, the option chromeOptions was renamed to goog:chromeOptions.

My code snippet started working as soon as I changed value of the constant `Facebook\WebDriver\Chrome\ChromeOptions::CAPABILITY` to `'goog:chromeOptions'`.

### Details

* Php-webdriver version: 1.5.0
* PHP version: 7.1.10
* Selenium server version: 3.8.1
* Operating system: Linux x86_64
* Browser used + version: Chrome 65

maidmaid:
A workaround could be to use `ChromeOptions::setExperimentalOption()`:

```php
$perfLoggingPrefs = new \stdClass();
$perfLoggingPrefs->enableNetwork = true;

$chromeOptions = new ChromeOptions();
$chromeOptions->setExperimentalOption('perfLoggingPrefs', $perfLoggingPrefs);
```

maidmaid:
Extra note following my previous comment : if your goal is to enable the [performance log](http://chromedriver.chromium.org/logging/performance-log), you need this extra instructions :

```php
$loggingPrefs = new \stdClass();
$loggingPrefs->performance = 'ALL';
$capabilities->setCapability('loggingPrefs', $loggingPrefs);
```
And then, for getting the logs :

```
$webDriver()->manage()->getLog('performance');
```

dotnetCarpenter:
@maidmaid Seems like a small error somehow in your code.


**ERROR**: _Facebook\WebDriver\Exception\ExpectedException: invalid argument: perfLoggingPrefs specified, but performance logging was not enabled_

```php
$perfLoggingPrefs = new \stdClass();
$perfLoggingPrefs->enableNetwork = true;

$chromeOptions = new ChromeOptions();
$chromeOptions->setExperimentalOption('perfLoggingPrefs', $perfLoggingPrefs);
$chromeOptions->addArguments(['--headless', '--no-sandbox', '--disable-gpu']);

$capabilities = DesiredCapabilities::chrome();
$capabilities->setCapability(ChromeOptions::CAPABILITY, $chromeOptions);
$capabilities->setCapability('logginPrefs', [
    'performance' => 'ALL',
    'browser' => 'ALL'
]);

[...]

PR #673, opened 2019-11-11
Make chromeOptions compatible with both JsonWire and W3C protocols

OndraM:
This workaround resolves issue which emerged after #560 .

Newer Selenium servers reject `chromeOptions` key in capabilities, because it does not contain vendor prefix, what is required as per W3C protocol. However simply updating the prefix would break on the other hand older Chromedrivers - and we want to maintain BC with version 1 of this library.

Ref. #652 #551 #573 #469


abdrmdn:
Thank

PR #677, opened 2019-11-14
Download a compatible version of chromedriver with chrome

Lctrs:
Following the [algorithm described on the official chromedriver website](https://chromedriver.chromium.org/downloads/version-selection) : 

> Here are the steps to select the version of ChromeDriver to download:
> 
> - First, find out which version of Chrome you are using. Let's say you have Chrome 72.0.3626.81.
> - Take the Chrome version number, remove the last part, and append the result to URL "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_". For example, with Chrome version 72.0.3626.81, you'd get a URL "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_72.0.3626".
> - Use the URL created in the last step to retrieve a small file containing the version of ChromeDriver to use. For example, the above URL will get your a file containing "72.0.3626.69". (The actual number may change in the future, of course.)
> - Use the version number retrieved from the previous step to construct the URL to download ChromeDriver. With version 72.0.3626.69, the URL would be "https://chromedriver.storage.googleapis.com/index.html?path=72.0.3626.69/".

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ We have a great community willing to help you!
 
 - **Via our Facebook Group** - If you have questions or are an active contributor consider joining our [facebook group](https://www.facebook.com/groups/phpwebdriver/) and contribute to communal discussion and support
 - **Via StackOverflow** - You can also [ask a question](https://stackoverflow.com/questions/ask?tags=php+selenium-webdriver) or find many already answered question on StackOverflow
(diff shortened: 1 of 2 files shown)


andrewnicols:
This helper may also be of use:
https://github.com/andrewnicols/chromedriver-wrapper

OndraM:
@Lctrs Merged to community branch manually, thanks!

PR #681, opened 2019-11-20
Switch to window takes a handle

andrewnicols:
https://www.w3.org/TR/webdriver/#switch-to-window

> Let handle be the result of getting the property "handle" from the parameters argument.

Apologies but I'm not quite sure how best to unit test this.

### DIFF
--- a/lib/Remote/RemoteTargetLocator.php
+++ b/lib/Remote/RemoteTargetLocator.php
@@ -89,7 +89,12 @@ class RemoteTargetLocator implements WebDriverTargetLocator
      */
     public function window($handle)
     {
-        $params = ['name' => (string) $handle];
+        if ($this->isW3cCompliant) {
+            $params = ['handle' => (string) $handle];
+        } else {
+            $params = ['name' => (string) $handle];
+        }
+
         $this->executor->execute(DriverCommand::SWITCH_TO_WINDOW, $params);
 
         return $this->driver;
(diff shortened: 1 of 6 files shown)


andrewnicols:
CLA now signed.

lib/Remote/RemoteTargetLocator.php
```diff
@@ -84,7 +84,7 @@ public function frame($frame)
      */
     public function window($handle)
     {
-        $params = ['name' => (string) $handle];
+        $params = ['handler' => (string) $handle];
```
OndraM: However keep in mind the compatibility with the old protocol must be kept - so there should be somewhere condition so that this is executed differently in JsonWire OSS protocol W3C WebDriver protocol. Have look at method `activeElement() ` in the same file, where something similar is done.

andrewnicols: Thanks @OndraM ,

I've used the same process as in activeElement.

I wonder if there's any merit to having it take `$isW3cCompliant` into the constructor like in RemoteMouse and doing that consistently across the board.

Let me know if you want me to make that change, or whether it's worth raising a separate issue for it.

andrewnicols:
As I've now come across a second function in the RemoteTargeLocator which needs to maintain b/c, I've passed the isW3cCompliant argument as an optional 3rd argument to the constructor.

This change is on a separate commit shared between this issue and #684

PR #682, opened 2019-11-21
Make ActiveElement endpoint W3C compatible

andrewnicols:
As defined in the specification 12.2.6:

HTTP Method   URI Template
GET           /session/{session id}/element/active

Source: https://www.w3.org/TR/webdriver/#get-active-element

PR #684, opened 2019-11-21
Frame switching W3C Compliance

andrewnicols:
Frame switching W3C Compliance

This change updates the frame functionality to comply with the W3C
specification.

Valid values of the `id` parameter are:
- null
- a Number between 0 and 2^16
- an Object representing a Web element

When the value is null, set the current browsing context to the current
top-level browsing context

When the value is a valid Number, let window be the associated window of
the current browsing context’s active document.

When the value is an Object, let element be the result of trying to get
a known element by web element reference id.

Source: https://www.w3.org/TR/webdriver/#switch-to-frame

I have also included a second commit, shared with #681 to pass isW3cCompliant to the constructor as we do in other classes.

### DIFF
--- a/lib/Remote/RemoteTargetLocator.php
+++ b/lib/Remote/RemoteTargetLocator.php
@@ -62,16 +62,39 @@ class RemoteTargetLocator implements WebDriverTargetLocator
     /**
      * Switch to the iframe by its id or name.
      *
-     * @param WebDriverElement|string $frame The WebDriverElement,
+     * @param WebDriverElement|null|int|string $frame The WebDriverElement,
      * the id or the name of the frame.
+     * When null, switch to the current top-level browsing context
+     * When int, switch to the WindowProxy identified by the value
+     * When an Element, switch to that Element.
+     *
+     * @throws \InvalidArgumentException
      * @return WebDriver The driver focused on the given frame.
      */
     public function frame($frame)
     {
-        if ($frame instanceof WebDriverElement) {
-            $id = ['ELEMENT' => $frame->getID()];
+        if ($this->isW3cCompliant) {
+            if ($frame instanceof WebDriverElement) {
+                $id = [JsonWireCompat::WEB_DRIVER_ELEMENT_IDENTIFIER => $frame->getID()];
+            } elseif ($frame === null) {
+                $id = null;
+            } elseif (is_int($frame)) {
+                $id = $frame;
+            } else {
+                throw new \InvalidArgumentException(
+                    'In W3C compliance mode frame must be either instance of WebDriverElement, integer or null'
+                );
+            }
         } else {
-            $id = (string) $frame;
+            if ($frame instanceof WebDriverElement) {
+                $id = ['ELEMENT' => $frame->getID()];
+            } elseif ($frame === null) {
+                $id = null;
+            } elseif (is_int($frame)) {
+                $id = $frame;
+            } else {
+                $id = (string) $frame;
+            }
         }
 
         $params = ['id' => $id];
--- a/tests/functional/RemoteTargetLocatorTest.php
+++ b/tests/functional/RemoteTargetLocatorTest.php
@@ -71,4 +71,76 @@ class RemoteTargetLocatorTest extends WebDriverTestCase
         $this->assertSame('input', $activeElement->getTagName());
         $this->assertSame('test_name', $activeElement->getAttribute('name'));
     }
+
+    /**
+     * @cover ::frame
+     */
+    public function testShouldSwitchToFrameByItsId()
+    {
+        $parentPage = 'This is the host page which contains an iFrame';
+        $firstChildFrame = 'This is the content of the iFrame';
+        $secondChildFrame = 'open new window';
+
+        $this->driver->get($this->getTestPageUrl('page_with_frame.html'));
+
+        $this->assertContains($parentPage, $this->driver->getPageSource());
+
+        $this->driver->switchTo()->frame(0);
+        $this->assertContains($firstChildFrame, $this->driver->getPageSource());
+
+        $this->driver->switchTo()->frame(null);
+        $this->assertContains($parentPage, $this->driver->getPageSource());
+
+        $this->driver->switchTo()->frame(1);
+        $this->assertContains($secondChildFrame, $this->driver->getPageSource());
+
+        $this->driver->switchTo()->frame(null);
+        $this->assertContains($parentPage, $this->driver->getPageSource());
+    }
+
+    /**
+     * @cover ::frame
+     */
+    public function testShouldSwitchToFrameByElement()
+    {
+        $this->driver->get($this->getTestPageUrl('page_with_frame.html'));
+
+        $element = $this->driver->findElement(WebDriverBy::id('iframe_content'));
+        $this->driver->switchTo()->frame($element);
+
+        $this->assertContains('This is the content of the iFrame', $this->driver->getPageSource());
+    }
+
+    /**
+     * @cover ::frame
+     * @group exclude-saucelabs
+     */
+    public function testShouldNotAcceptStringAsFrameIdInW3cMode()
+    {
+        self::skipForJsonWireProtocol();
+
+        $this->driver->get($this->getTestPageUrl('page_with_frame.html'));
+
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionMessage(
+            'In W3C compliance mode frame must be either instance of WebDriverElement, integer or null'
+        );
+
+        $this->driver->switchTo()->frame('iframe_content');
+    }
+
+    /**
+     * @cover ::frame
+     * @group exclude-saucelabs
+     */
+    public function testShouldAcceptStringAsFrameIdInJsonWireMode()
+    {
+        self::skipForW3cProtocol();
+
+        $this->driver->get($this->getTestPageUrl('page_with_frame.html'));
+
+        $this->driver->switchTo()->frame('iframe_content');
+
+        $this->assertContains('This is the content of the iFrame', $this->driver->getPageSource());
+    }
 }
--- a/tests/functional/WebDriverTestCase.php
+++ b/tests/functional/WebDriverTestCase.php
@@ -126,7 +126,7 @@ class WebDriverTestCase extends TestCase
     public static function skipForJsonWireProtocol($message = 'Not supported by JsonWire protocol')
     {
         if (getenv('GECKODRIVER') !== '1'
-            && (getenv('CHROMEDRIVER') !== '1' || getenv('DISABLE_W3C_PROTOCOL') === '1')) {
+            && (getenv('BROWSER_NAME') !== 'chrome' || getenv('DISABLE_W3C_PROTOCOL') === '1')) {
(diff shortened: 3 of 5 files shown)


lib/Remote/RemoteTargetLocator.php
```diff
@@ -63,10 +68,20 @@ public function defaultContent()
      */
     public function frame($frame)
     {
-        if ($frame instanceof WebDriverElement) {
-            $id = ['ELEMENT' => $frame->getID()];
+        if ($this->isW3cCompliant) {
+            if ($frame instanceof WebDriverElement) {
+                $id = [JsonWireCompat::WEB_DRIVER_ELEMENT_IDENTIFIER => $frame->getID()];
+            } else if (null === $frame) {
```
OndraM:
```suggestion
            } elseif ($frame === null) {
```

Please don't use yoda conditions, they are harder to read :).

andrewnicols:
Hi @OndraM ,

I've realised that we need to produce a better error message when a string is passed. Right now if you pass a string as $frame, then it's cast to an int, specifically 0, and passes.

Not sure whether we should add a new Exception here or not?

The error returned by W3C WebDriver is:
```
Facebook\WebDriver\Exception\UnrecognizedExceptionException: invalid argument: 'id' can not be string
```

Sadly we don't have any kind of InvalidArgument exception or similar.

What are your thoughts?

I've added some unit tests and detect the current `Facebook\WebDriver\Exception\UnrecognizedExceptionException` but should be easy to change.

OndraM:
@andrewnicols It seems "invalid argument" error is missing in W3C exceptions handling 🤔.

Could you in this pull request add unit tests using `UnrecognizedExceptionException` and `expectExceptionMessage`, and I will have look at the missing exception :)?

OndraM:
@andrewnicols See #686 with all the new exceptions :ok_hand: .

You can now rebase your branch on top of `w3c-exceptions` branch and change base of this PR to `w3c-exceptions` - then you could already use the new exception.

andrewnicols:
Thanks @OndraM ,

All rebased.
Just thinking - what's the best way of excluding non-W3C browsers from the tests which will only work on W3C browsers? I.e. it's valid to use a string with a JSONWire browser but not with W3C. Do we have any ability to use any `@group` notation?

OndraM:
@andrewnicols You can use `self::skipForJsonWireProtocol()` method to skip those tests.

andrewnicols:
Ah thanks. I’ll do that when I get home in a couple of hours.

andrewnicols:
Right, Unit tests added and skipping the string unit test for JSONWire protocol

andrewnicols:
Note: This is based on #682 as it uses the activeElement functionality in the unit tests.

andrewnicols:
Rebased on community.

andrewnicols:
Rebased again:
* fixed missing `use` for exception
* corrected missing `is_numeric` check
* changed `assertEquals` to `assertContains` for content of the frame checks

andrewnicols:
And force pushed one final time because I had the order of the `use` statements wrong.

I'm going to stop changing things now!

OndraM:
I rebased the branch - there were conflicts following merger of #682

lib/Remote/RemoteTargetLocator.php
```diff
@@ -68,10 +68,22 @@ public function defaultContent()
      */
     public function frame($frame)
     {
-        if ($frame instanceof WebDriverElement) {
-            $id = ['ELEMENT' => $frame->getID()];
+        if ($this->isW3cCompliant) {
+            if ($frame instanceof WebDriverElement) {
+                $id = [JsonWireCompat::WEB_DRIVER_ELEMENT_IDENTIFIER => $frame->getID()];
+            } else if ($frame === null) {
```
OndraM:
```suggestion
            } elseif ($frame === null) {
```

lib/Remote/RemoteTargetLocator.php
```diff
@@ -68,10 +68,22 @@ public function defaultContent()
...
+                $id = null;
+            } else if (is_numeric($frame)) {
```
OndraM:
```suggestion
            } elseif (is_numeric($frame)) {
```

lib/Remote/RemoteTargetLocator.php
```diff
@@ -68,10 +68,22 @@ public function defaultContent()
      */
```
OndraM: Null should be mentioned in `@param` - and also with explanation what each value means (WebDriverElement/null/numeric id)

lib/Remote/RemoteTargetLocator.php
```diff
@@ -68,10 +68,22 @@ public function defaultContent()
...
-            $id = ['ELEMENT' => $frame->getID()];
+        if ($this->isW3cCompliant) {
+            if ($frame instanceof WebDriverElement) {
+                $id = [JsonWireCompat::WEB_DRIVER_ELEMENT_IDENTIFIER => $frame->getID()];
+            } else if ($frame === null) {
+                $id = null;
+            } else if (is_numeric($frame)) {
+                $id = (int) $frame;
+            } else {
+                $id = (string) $frame;
```
OndraM: If W3C only allows $frame to be either `WebDriverElement`, `null` or numeric ID, why should we accept string (and send it to the driver)?

tests/functional/RemoteTargetLocatorTest.php
```diff
@@ -71,4 +72,86 @@ public function testActiveElement()
...
+                $otherChildFrame,
+            ],
+            'int 1' => [
+                1,
+                $otherChildFrame,
+            ],
+            'floa 1' => [
+                1.0,
+                $otherChildFrame,
+            ],
```
OndraM: I would prefer having datasets formatted on one row if they fits, like this:
```php
 'null' => [null, $parentFrame],
```

tests/functional/RemoteTargetLocatorTest.php
```diff
@@ -71,4 +72,86 @@ public function testActiveElement()
         $this->assertSame('input', $activeElement->getTagName());
         $this->assertSame('test_name', $activeElement->getAttribute('name'));
     }
+
+    /**
+     * Data provider for frame().
+     */
```
OndraM: The comment is unnecessary

andrewnicols:
It looks like there may be some unit test fails.
I may have misread the inplementation of the numeric frame IDs. The Webdriver spec is particularly unclear on the meaning of them (See https://github.com/w3c/webdriver/issues/1205 for example).

It’s late here (Perth, Australia), so I won’t be able to look til tomorrrow unless you find a better solution before then.

Thanks

tests/functional/RemoteTargetLocatorTest.php
```diff
@@ -71,4 +72,86 @@ public function testActiveElement()
...
+    public function frameProvider()
```
OndraM: I would prefer this name (I will later rename all other providers to similar naming scheme `provide...()`)

```suggestion
    public function provideFrame()
```

tests/functional/RemoteTargetLocatorTest.php
```diff
@@ -71,4 +72,86 @@ public function testActiveElement()
...
+                $otherChildFrame,
+            ],
+        ];
+    }
+
+    /**
+     * @cover ::frame
+     * @dataProvider frameProvider
+     */
+    public function testValidFrames($frame, $content)
```
OndraM:
```suggestion
    public function testValidFrames($frame, $expectedContent)
```

tests/functional/RemoteTargetLocatorTest.php
```diff
@@ -71,4 +72,86 @@ public function testActiveElement()
...
+    {
+        $this->driver->get($this->getTestPageUrl('page_with_frame.html'));
+        $this->driver->switchTo()->frame($frame);
+        $this->assertContains(
+            $content,
+            $this->driver->switchTo()->activeElement()->getText()
+        );
```
OndraM: I suggest separating every part of the arrange/act/assert phase of the test by new line. In this case it means newline after eachline:
```
        $this->driver->get($this->getTestPageUrl('page_with_frame.html'));

        $this->driver->switchTo()->frame($frame);

        $this->assertContains(
            $content,
            $this->driver->switchTo()->activeElement()->getText()
        );
```

tests/functional/RemoteTargetLocatorTest.php
```diff
@@ -71,4 +72,86 @@ public function testActiveElement()
...
+     * @cover ::frame
+     * @dataProvider frameProvider
+     */
+    public function testValidFrames($frame, $content)
+    {
+        $this->driver->get($this->getTestPageUrl('page_with_frame.html'));
+        $this->driver->switchTo()->frame($frame);
+        $this->assertContains(
+            $content,
+            $this->driver->switchTo()->activeElement()->getText()
```
OndraM: Do we need to use `switchTo()->activeElement()`? Wouldn't something like `$this->driver->getPageSource()` do the trick? (also in the other tests)

lib/Remote/RemoteTargetLocator.php
andrewnicols: So do you think this should except? The only reason I didn’t do so was because I wasn’t sure what exception I should throw. I haven’t seen any/many examples where we throw an exception rather than translate an error into an exception.

Should we just throw the same exception that would be thrown anyway?

andrewnicols:
I’ll make those style changes tomorrow. Is there a specific style guide to follow for future reference so I don’t waste your time in future?

OndraM:
@andrewnicols Something is covered by codestyle check... Some other rules are not yet, sorry :-/. I plan to improve to codestyle check to contain more rules - but this will probably came with 2.0.

lib/Remote/RemoteTargetLocator.php
OndraM: We can just throw `\InvalidArgumentException` (the PHP's InvalidArgumentException, not the one from WebDriver). We already do it elsewhere in similar case.

tests/functional/RemoteTargetLocatorTest.php
andrewnicols: I've removed some of these test cases as they were invalid.
We should not accept string or floats and cast to int because the frame number comes from a Number and it is perfectly valid to have a frame whose name is a stringy int.

I've added new tests for stringy integers.

I also detected an existing bug with the JSONWire implementation where it did not support being passed null.
I've confirmed that the following are supported in JSONWire (https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidframe):
> id - {string|number|null|WebElement JSON Object} Identifier for the frame to change focus to.

tests/functional/RemoteTargetLocatorTest.php
andrewnicols: Switched to getPageSource.

OndraM:
Nice @andrewnicols! I've done few changes in the tests, especially removed the data providers, because for every dataset in the provided new browser instance must be started - which is very time expensive. So it is generally better to avoid data provides in functional tests.

Otherwise looks good to merge :+1: , I will do so after CI builds finises. Nice work!

PR #686, opened 2019-11-24
Add all W3C WebDriver exceptions

OndraM:
Implements all W3C WebDriver exceptions according to https://w3c.github.io/webdriver/#errors.

Those which didn't make it to the new standard are marked as deprecated.

Ref. #469


OndraM:
Failed builds are not related to this PR.

PR #687, opened 2019-11-25
Correct mouseUp W3C Action

andrewnicols:
I just discovered that the mouseUp MouseAction was actually calling pointerDown instead of pointerUp.

Fixed and provided with a new unit test.


andrewnicols:
Note: With this, and the other related pull requests that I've raised, I now have passing unit tests for a new Facebook Webdriver Mink Driver.

OndraM:
Nice catch, thanks @andrewnicols!

PR #693, opened 2019-11-28
Allow specification of firefox options for non-gecko

andrewnicols:
Note: This just allows us to specify options like headless.
I have not:
* fixed the `skipForJsonWireProtocol` issue . I think that we need a non-static version of these functions which checks the W3C attribute of the driver instance.
* added any new options. I was tempted to add `['log' => ['level' => 'trace']]` to aid debugging. Currently I"m adding that manually

### DIFF
--- a/tests/functional/WebDriverTestCase.php
+++ b/tests/functional/WebDriverTestCase.php
@@ -63,8 +63,10 @@ class WebDriverTestCase extends TestCase
                 }
 
                 $this->desiredCapabilities->setCapability(ChromeOptions::CAPABILITY, $chromeOptions);
-            } elseif (getenv('GECKODRIVER') === '1') {
-                $this->serverUrl = 'http://localhost:4444';
+            } elseif ($browserName === WebDriverBrowserType::FIREFOX) {
+                if (getenv('GECKODRIVER') === '1') {
+                    $this->serverUrl = 'http://localhost:4444';
+                }
                 $this->desiredCapabilities->setCapability(
                     'moz:firefoxOptions',
                     ['args' => ['-headless']]


OndraM:
What is the use-case for this change? In current travis build set this does not make a difference (only Firefox browser capable of headless mode is the one using Geckodriver).

williamdes:
Fro the code I can see I would say that you can enter the if (gecko) only if you are not using `getenv('GECKODRIVER') === '1')`

In conclusion actual code seems to deny access to capabilities if you have `getenv('GECKODRIVER') === '1')`

andrewnicols:
> What is the use-case for this change? In current travis build set this does not make a difference (only Firefox browser capable of headless mode is the one using Geckodriver).

The use-case is running unit tests for Firefox through Selenium headless, which would be supported with this change and means that I can use the Selenium Docker images for unit testing without editing the config.

i.e. I will be able to do:
```
docker run -d --rm --name ff -p 4444:4444 selenium/standalone-firefox:3
FIXTURES_HOST='http://host.docker.internal:8000' GECKODRIVER=0 BROWSER_NAME=firefox ./vendor/bin/phpunit tests/functional/RemoteKeyboardTest.php
```

This allows much easier testing of multiple versions of Firefox/Selenium via the Selenium docker images.

I'd like to push another PR for specifying the `serverUrl` (WD URL) so that it's possible to run multiple wd hosts for local testing without having to kill them between each run but I'll keep that separate.

OndraM:
@andrewnicols Oh, I see, thats a nice workflow. :+1: 

I'm gonna merge this , but do you think it would be possible to expand CONTRIBUTING.md to include some information how to run tests against Docker image like you do? (Or maybe we can define some script to composer.json to make it even easier)?.

PR #694, opened 2019-11-28
Allow specification of alternate host for fixtures

andrewnicols:
To allow me to test using docker images, I have to manually edit the `getTestPgeUrl` function in the Driver testcase.

This is just a small change to allow the fixtures host to be specified as an environment variable called `FIXTURES_HOST`.

### DIFF
--- a/tests/functional/WebDriverTestCase.php
+++ b/tests/functional/WebDriverTestCase.php
@@ -139,7 +139,12 @@ class WebDriverTestCase extends TestCase
      */
     protected function getTestPageUrl($path)
     {
-        return 'http://localhost:8000/' . $path;
+        $host = 'http://localhost:8000';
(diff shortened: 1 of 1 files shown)


tests/functional/WebDriverTestCase.php
```diff
@@ -139,7 +139,12 @@ public static function skipForJsonWireProtocol($message = 'Not supported by Json
      */
     protected function getTestPageUrl($path)
     {
-        return 'http://localhost:8000/' . $path;
+        $host = 'http://localhost:8000';
+        if ($alternateHost = getenv('FIXTURES_HOST')) {
+            $host = $alternateHost;
+        }
+
+        return "{$host}/{$path}";
```
OndraM: We usually use this style of concatenation through the codebase:

```suggestion
        return $host . '/' . $path;
```

I will update it and merge the PR :)

PR #696, opened 2019-11-30
Change environment variable defining path to ChromeDriver executable

OndraM:
Replaces #535

### DIFF
--- a/lib/Chrome/ChromeDriverService.php
+++ b/lib/Chrome/ChromeDriverService.php
@@ -19,19 +19,23 @@ use Facebook\WebDriver\Remote\Service\DriverService;
 
 class ChromeDriverService extends DriverService
 {
-    // The environment variable storing the path to the chrome driver executable.
+    /**
+     * The environment variable storing the path to the chrome driver executable.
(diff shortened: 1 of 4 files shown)


lib/Chrome/ChromeDriverService.php
```diff
@@ -19,19 +19,23 @@
...
+    // The environment variable storing the path to the chrome driver executable
+    const CHROME_DRIVER_EXECUTABLE = 'WEBDRIVER_CHROME_DRIVER';
 
     /**
      * @return static
      */
     public static function createDefaultService()
     {
-        $exe = getenv(self::CHROME_DRIVER_EXE_PROPERTY);
+        $exe = getenv(self::CHROME_DRIVER_EXE_PROPERTY) ?: getenv(self::CHROME_DRIVER_EXECUTABLE);
```
andrewnicols: This will give priority to the old value.
Given the old version is deprecated, shouldn't this be swapped, i.e.

```
    $exe = getenv(self::CHROME_DRIVER_EXECUTABLE) ?: getenv(self::CHROME_DRIVER_EXE_PROPERTY);
```

PR #697, opened 2019-12-01
Maintain BC for renamed TimeOutException

Lctrs:


OndraM:
Hi, the class was just renamed from `TimeOutException` to `TimeoutException`. And as class names are case insensitive in PHP, I assumed this is a safe change. (And I was more afraid about what would happen if we have both the versions in case inensitive filesystems like Windows and Mac).

But I may be wrong - could you please provide example how exactly is this causing BC break?

Lctrs:
Static code analysis tools suck as PHPStan will break because they are case sensitive.

Also libraries and apps using PSR-0/4 compatible loader (such as the one generated by Composer) will break because they will expect to find the class in `TimeOutException.php` and not `TimeoutException.php`.

OndraM:
> Static code analysis tools such as PHPStan will break because they are case sensitive.

But it won't break the code. There are many other things we could do, which will not be a BC break, but will cause static analysis of user library to complain (eg. add some missing typehints atd.).

> Also PSR-0/4 compatible loaders (such as the one generated by Composer) will break because they will expect to find the class in `TimeOutException.php` and not `TimeoutException.php`.

How so? If you update dependencies,autoloading is regenerated...

Also I'm still unsure what happens on mac/win if you have two files with the same name in the repoisitory and install the repository via composer. Which one of the two files gets installed?

Lctrs:
Take the following class : 

```php
<?php

use Facebook\WebDriver\Exception\TimeOutException;
use Facebook\WebDriver\JavaScriptExecutor;
use Facebook\WebDriver\WebDriver;

class A
{
    /** @var WebDriver */
    private $driver;

    public function wait($condition) : bool
    {
        $wait = $this->driver->wait();

        try {
            return (bool) $wait->until(static function (JavaScriptExecutor $driver) {
                throw new TimeOutException('A timeout');
            });
        } catch (TimeOutException $e) {
            return false;
        }
    }
}
```

Composer will try to autoload the class `TimeOutException` from `TimeOutException.php`. Since the name of the file has changed, it won't be able to load the class and thus existing code is broken. That's why even changing case of a class is a BC Break.

But with the current state of this PR, it won't work either. the new `TimeoutException` must extends the now-deprecated `TimeOutException` (and not the other way around like in this PR) so that existing script will also catch `TimeoutException`.

I haven't had the time to try on Windows and Mac yet.

Lctrs:
@OndraM Here is an alternative proposal that will be compatible on all platform.

Lctrs:
In fact, always loading TimeoutException is enough :)

OndraM:
That is nice solution, thanks @Lctrs!

PR #695, opened 2019-11-29
Accept W3C capabilitites from the browser and provide them as OSS ones

OndraM:
Fixes #692

PR #703, opened 2019-12-04
Remove platform ANY from required capabilities

andrewnicols:
The 'ANY' platform is a pseudo-platform and therefore cannot be used in any kind of required capability, therefore it must be filtered out when specifying the firstMatch or alwaysMatch capability sets. Attempting to use it in either of these leads to an error: "Unable to find a matching set of capabilities".

I'm not sure whether this is the _best_ approach to take here. I feel that, ideally, we should have a RequiredCapabilites class which extends DesiredCapabilities and can clone from it, but I think that also poses other issues too. Happy to take direction on this one.

Note: This will likely conflict with #695. Happy to rebase this once that issue lands.

### DIFF
--- a/lib/Remote/DesiredCapabilities.php
+++ b/lib/Remote/DesiredCapabilities.php
@@ -199,9 +199,10 @@ class DesiredCapabilities implements WebDriverCapabilities
     }
 
     /**
+     * @param bool $forRequiredCapabilities
      * @return array
      */
-    public function toW3cCompatibleArray()
+    public function toW3cCompatibleArray($forRequiredCapabilities = false)
     {
         $allowedW3cCapabilities = [
             'browserName',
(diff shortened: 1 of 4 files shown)


OndraM:
Sorry, I didn't have time to look into this so far... But let me just ask - Is this something we necessarily need for 1.8.0 release?

andrewnicols:
We probably don't _need_ this for 1.8.0.

OndraM:
Nah, we actually might need this, because otherwise you are not able to start Firefox/Geckodriver using:

`$driver = RemoteWebDriver::create($host,$DesiredCapabilities::firefox());`

OndraM:
I attempted to do this in #703 - what do you think @andrewnicols? Will it work that way?

OndraM:
Fixed by #703

PR #707, opened 2019-12-07
Use clear endpoint to clear element content even in W3C mode

### LINKED ISSUE #704: Calling "clear()" before "sendKeys()" causes some chars to be eaten
Hello,

I could dig some more the issue from yesterday, and reproduce it.

I'm trying to log in https://the-internet.herokuapp.com/login
I'm using the following code, which reproduces the way Symfony Panther calls webdriver:
```php
$email = $driver->findElement(WebDriverBy::name('username'));
$email->clear();
$email->sendKeys('tomsmith');

$password = $driver->findElement(WebDriverBy::name('password'));
$password->clear();
$password->sendKeys('SuperSecretPassword!');

$driver->findElement(WebDriverBy::cssSelector('button[type="submit"]'))->click();
```

Last character from `username` input is never typed, causing login to fail.

The issue disappear if `$email->clear();` and `$password->clear();` are commented.

It is not a timing issue, adding a `sleep(5)` or so does not help.
The crash looks to be caused by [lines 76-93 of RemoteWebElement::clear()](https://github.com/facebook/php-webdriver/blob/community/lib/Remote/RemoteWebElement.php#L76). Commenting those or replacing "WebDriverKeys::BACKSPACE" by "WebDriverKeys::DELETE" fixed it.

Here's a minimal gist to reproduce the issue: https://gist.github.com/romaricdrigon/1c2c3d3ab99644f61ba21f8334d6fc18

### Details

* Php-webdriver version: latest `community` (`1.8.x-dev`), it does not happen in `1.7.1`
* PHP version: 7.4.0
* Selenium server version: using ChromeDriver directly, 78.0.3904.70 or 78.0.3904.105
* Operating system: reproduced both on Debian (dockerized) and macOS
* Browser used + version: Chromium 78.0.3904.108 or Chrome 78.0.3904.108

OndraM:
This _may_ fix #704 and also resolve the issue mentioned in https://github.com/facebook/php-webdriver/pull/691#issuecomment-559883408 . 

@dunglas We are not sure why the actions sequence was there in the first place, do you remember why you add it? (Test pass even without it.)

Cc @andrewnicols @romaricdrigon - could you guys try, if it fixes the issues for you? 
`composer require facebook/webdriver:dev-w3c-clear@dev`

### DIFF
--- a/lib/Remote/RemoteWebElement.php
+++ b/lib/Remote/RemoteWebElement.php
@@ -62,7 +62,7 @@ class RemoteWebElement implements WebDriverElement, WebDriverLocatable
     }
 
     /**
-     * If this element is a TEXTAREA or text INPUT element, this will clear the value.
+     * Clear content editable or resettable element
      *
      * @return RemoteWebElement The current instance.
      */
@@ -72,25 +72,6 @@ class RemoteWebElement implements WebDriverElement, WebDriverLocatable
             DriverCommand::CLEAR_ELEMENT,
             [':id' => $this->id]
         );
-
-        if ($this->isW3cCompliant) {
-            $this->executor->execute(DriverCommand::ACTIONS, [
-                'actions' => [
-                    [
-                        'type' => 'key',
-                        'id' => 'keyboard',
-                        'actions' => [
-                            ['type' => 'keyDown', 'value' => WebDriverKeys::CONTROL],
-                            ['type' => 'keyDown', 'value' => 'a'],
-                            ['type' => 'keyUp', 'value' => WebDriverKeys::CONTROL],
-                            ['type' => 'keyUp', 'value' => 'a'],
-                            ['type' => 'keyDown', 'value' => WebDriverKeys::BACKSPACE],
-                            ['type' => 'keyUp', 'value' => WebDriverKeys::BACKSPACE],
-                        ],
-                    ],
-                ],
-            ]);
-        }
 
         return $this;
     }


romaricdrigon:
Hello,

It fixes the issue for me. Though now I see another issue appear on some edge case, I'm working on reproducing it.

OndraM:
@romaricdrigon Thanks! So I will wait with merging this, to make sure it won't break anything :)

romaricdrigon:
Thanks, I update my previous post!

andrewnicols:
Thanks @OndraM ,

Looks good to me. Clears them without issue and the backspace issue seems to be gone. The original issue only seems to have affected Chrome.

romaricdrigon:
I investigated, the small issue I mentioned earlier is unrelated to this PR (and actually, I think is was a bug in 1.7). So 👍 for this, thank you @OndraM for working on it!

OndraM:
Great, thanks for the feedback, merging now :ok_hand:

jedwards1211:
I don't use this library, but I wanted to give y'all a heads up that the clear element command isn't a realistic simulation of user behavior, and because it doesn't fire an `input` event, it doesn't work properly in some cases, like React controlled inputs with react-final-form, and one has to use a workaround like Ctrl-A/Command-A Backspace in that case.

OndraM:
@jedwards1211 Thanks - though the proper behavior is related to W3C WebDriver standard. So for anyone interested in this topic, there is discussion about this in W3C WebDriver project: https://github.com/w3c/webdriver/issues/1630

PR #708, opened 2019-12-08
Add PHP 7.4 Travis build

OndraM:
Also replace mock-objects with PHP 5.6-7.4 compatible fork sminnee/phpunit-mock-objects


OndraM:
We can change to `7.4` instead of `7.4snapshot` only after `https://travis-ci.community/t/some-extensions-are-missing-in-php-7-4-0-zip-gmp-sodium/6320` is fixed

PR #713, opened 2019-12-15
Update .travis.yml

andreybolonin:


OndraM:
Hi @andreybolonin , thanks for the contribution, however 7.4 build on Travis is still broken and we cannot use it, see https://github.com/facebook/php-webdriver/pull/708#issuecomment-563214676 .

Issue #683, opened 2019-11-21
Missing commands from W3C specification


andrewnicols:
I've just been through the spec document comparing to the actions in the HttpCommandExecutor. The following commands defined in the W3C specification are currently missing:

- [x] 11.7 Switch to Parent Frame #718
`POST  /session/{session id}/frame/parent`
- [x] 11.8.4 Minimize Window #719
`POST  /session/{session id}/window/minimize`
- [x] 11.8.5 Fullscreen Window #720
`POST  /session/{session id}/window/fullscreen`
- [x] 14.2 Get Named Cookie #717 
`GET  /session/{session id}/cookie/{name}`
- [ ] 15.6 Release Actions #755 
`DELETE  /session/{session id}/actions`
- [x] 17.2 Take Element Screenshot #701 
`GET  /session/{session id}/element/{element id}/screenshot`

Source: https://www.w3.org/TR/webdriver/
Ref. #469

OndraM:
@andrewnicols Thanks for the overview. Do you have an idea about the opposite case - ie. commands which were implemented in JsonWire OSS protocol and are not part of W3C specification?

andrewnicols:
Thanks @OndraM ,

I guess that depends on the way we want to approach things generally.

You've started to mark JSONWire-only exceptions as deprecated, so I'd suggest that's would be most consistent but I assume they'll be deprecated and never removed to support testing of older non-W3C browsers.

I feel an exception is the only sane way to go.

OndraM:
You are right, nothing won't be removed in php-webdriver 1.x.

However they will be removed in php-webdriver 2.0, as it will support only W3C WebDriver. See #657 for ongoing discussion about scope php-webdriver 2.0.


The reason I marked those exceptions deprecated is to announce users this is something they will need to change in the future (ie. they depend on behavior which will be removed in next major version). Also by marking them deprecated, it will be easier to distinguish what should be removed once we create branch for 2.0 :).

andrewnicols:
Perhaps a new Exception for `W3CNotImplemented` or `NotImplementedbyDriver` for those then?

OndraM:
@andrewnicols I'm not sure I understand - where would you use those exceptions?

OndraM:
@andrewnicols BTW take element screenshot is already implemented here - do you think we can use this PR? https://github.com/facebook/php-webdriver/pull/515

andrewnicols:
[...]

PR #718, opened 2019-12-22
Add switch to parent frame command

andrewnicols:
Part of #683

### DIFF
--- a/lib/Remote/RemoteTargetLocator.php
+++ b/lib/Remote/RemoteTargetLocator.php
@@ -98,6 +98,18 @@ class RemoteTargetLocator implements WebDriverTargetLocator
     }
 
     /**
+     * Switch to the parent iframe.
+     *
+     * @return WebDriver The driver focused on the given frame.
+     */
+    public function parent()
+    {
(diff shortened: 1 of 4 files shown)


lib/WebDriverTargetLocator.php
```diff
@@ -37,6 +37,13 @@ public function defaultContent();
      */
     public function frame($frame);
 
+    /**
+     * Switch to the parent iframe.
+     *
+     * @return WebDriver The driver focused on the given frame.
```
OndraM: Add something like `@todo Add in next major release` - so that we find if when preparing 2.0.

lib/Remote/RemoteTargetLocator.php
```diff
@@ -97,6 +97,19 @@ public function frame($frame)
...
+     * Switch to the parent iframe.
+     *
+     * @return WebDriver The driver focused on the given frame.
+     */
+    public function parent()
+    {
+        $this->executor->execute(DriverCommand::SWITCH_TO_PARENT_FRAME, []);
+
+        return $this->driver;
+
```
OndraM:
```suggestion
```

PR #719, opened 2019-12-22
Minimize window

andrewnicols:
Add the minimizeWindow command:
11.8.4 Minimize Window
`POST  /session/{session id}/window/minimize`

Note: There is a bug in chrome/chromedriver which prevents the visibilityState from being checked in headless mode. I've reported that in https://bugs.chromium.org/p/chromedriver/issues/detail?id=3275 and have added it to a new exclude-group for phpunit.

### DIFF
--- a/lib/WebDriverWindow.php
+++ b/lib/WebDriverWindow.php
@@ -16,6 +16,7 @@
 namespace Facebook\WebDriver;
 
 use Facebook\WebDriver\Exception\IndexOutOfBoundsException;
+use Facebook\WebDriver\Exception\UnsupportedOperationException;
 use Facebook\WebDriver\Remote\DriverCommand;
 use Facebook\WebDriver\Remote\ExecuteMethod;
 
@@ -28,10 +29,15 @@ class WebDriverWindow
      * @var ExecuteMethod
      */
     protected $executor;
+    /**
+     * @var bool
+     */
+    protected $isW3cCompliant;
 
-    public function __construct(ExecuteMethod $executor)
+    public function __construct(ExecuteMethod $executor, $isW3cCompliant = false)
     {
         $this->executor = $executor;
+        $this->isW3cCompliant = $isW3cCompliant;
     }
 
     /**
@@ -70,6 +76,22 @@ class WebDriverWindow
             $size['width'],
             $size['height']
         );
+    }
+
+    /**
+     * Minimizes the current window if it is not already minimized.
+     *
+     * @return WebDriverWindow The instance.
+     */
+    public function minimize()
+    {
+        if (!$this->isW3cCompliant) {
+            throw new UnsupportedOperationException('Minimize window is only supported in W3C mode');
+        }
+
+        $this->executor->execute(DriverCommand::MINIMIZE_WINDOW, []);
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


lib/WebDriverWindow.php
```diff
@@ -72,6 +77,20 @@ public function getSize()
...
+    /**
+     * Minimizes the current window if it is not already maximized
+     *
+     * @return WebDriverWindow The instance.
+     */
+    public function minimize()
+    {
+        if ($this->isW3cCompliant) {
+            $this->executor->execute(DriverCommand::MINIMIZE_WINDOW, []);
+        }
```
OndraM: As minimize is W3C specific, shouldn't we throw an exception if it is called in non-w3c mode?

```php
if (!$this->isW3cCompliant) {
    throw new UnsupportedOperationException('Minimize window is only supported in W3C mode');`
}

$this->executor->execute(DriverCommand::MINIMIZE_WINDOW, []);
```

tests/functional/WebDriverWindowTest.php
```diff
@@ -48,6 +48,20 @@ public function testShouldMaximizeWindow()
         $this->assertGreaterThanOrEqual($sizeBefore->getHeight(), $sizeAfter->getHeight());
     }
 
+    /**
+     * @group exclude-chrome
+     */
+    public function testShouldMinimizeWindow()
+    {
+        $this->assertSame('visible', $this->driver->executeScript('return document.visibilityState;'));
```
OndraM: The test should also by skipped in JsonWire mode. This will also make it being skipped in HtmlUnit, as it is not W3C compatible afaik.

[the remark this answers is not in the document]
andrewnicols: I adjusted this to skip for JsonWire as suggested.

andrewnicols: Also added a JsonWire-only test to assert the fail.

tests/functional/WebDriverWindowTest.php
```diff
@@ -48,6 +50,34 @@ public function testShouldMaximizeWindow()
...
+
+        $this->driver->manage()
+            ->window()
+            ->minimize();
+
+        $this->assertSame('hidden', $this->driver->executeScript('return document.visibilityState;'));
+    }
+
+    public function testShouldMinimizeWindowJsonWire()
```
OndraM: I don't think we need this kind of tests... They IMHO don't add a real value, only slow down the build - there is no interaction with the browser, dispite this the browser is started which takes eg. on SauceLabs 10+ seconds. 

So this tests which only tests internal logic could be unit tests, however, I think it is unnecessary altogether (considering we will ditch JsonWire support later).

andrewnicols:
Okay, addressed that last one. My main work is on a project where unit tests are required for those kinds of change. Hard habit to break.

tests/functional/WebDriverWindowTest.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Facebook\WebDriver;
 
+use Facebook\WebDriver\Exception\UnsupportedOperationException;
```
OndraM: Unused import (see codestyle build)

andrewnicols:
Should be good to go now.

PR #720, opened 2019-12-22
Fullscreen window

andrewnicols:
I've included an extra commit here because the W3C mode doesn't take any arguments but JSONWire does but we were still including the args still.

### DIFF
--- a/lib/WebDriverWindow.php
+++ b/lib/WebDriverWindow.php
@@ -101,10 +101,14 @@ class WebDriverWindow
      */
     public function maximize()
     {
-        $this->executor->execute(
-            DriverCommand::MAXIMIZE_WINDOW,
-            [':windowHandle' => 'current']
-        );
+        if ($this->isW3cCompliant) {
+            $this->executor->execute(DriverCommand::MAXIMIZE_WINDOW, []);
+        } else {
+            $this->executor->execute(
+                DriverCommand::MAXIMIZE_WINDOW,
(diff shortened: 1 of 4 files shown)


andrewnicols:
I think that this should be ready to land. I've had to exclude edge for now because SauceLabs seems to be running an old version (their docs suggest EdgeHTML 18) which does not support the fullscreen command.

Microsoft are now releasing Edge based on Blink so we should be able to remove this exclusion when SauceLabs bump their edge versions.

lib/WebDriverWindow.php
```diff
@@ -79,10 +84,28 @@ public function getSize()
...
+    }
+
+    /**
+     * Makes the current window full screen.
+     *
+     * @return WebDriverWindow The instance.
+     */
+    public function fullscreen()
+    {
+        if ($this->isW3cCompliant) {
```
OndraM: The fullscreen was not part of JsonWire (but may work eg. in Chrome), so I think we should either:
a) execute the fullscreen command regaredless the protocol (and let i fail by exception on the remote end)
b) throw exception if not in w3c mode (like we do in `minimize()`)

But silently doing nothing is IMO not the best approach.

andrewnicols:
Thanks,

That makes total sense. I've changed it to throw an exception and have modified the test to skip for the JsonWire protocol.

OndraM:
Thanks @andrewnicols!

PR #726, opened 2019-12-26
Migrate codestyle check to GitHub actions

OndraM:
To lower the stress on Travis and speed up the build a bit.


.github/workflows/php.yaml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+  schedule:
+    - cron:  '0 3 * * *'
+
+jobs:
+  analyze:
+      name: "Code style and static analysis"
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v1
```
Lctrs:
```suggestion
        - uses: actions/checkout@v2
```

.github/workflows/php.yaml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v1
+        with:
+          php-version: '7.4'
+          extensions: mbstring, intl, zip
+
+      - name: Install dependencies
```
williamdes: you could add a cache section to speed up composer ?

OndraM: :+1: True, we may improve it in the future, but first lets get Actions working at least somehow :)

PR #732, opened 2020-01-12
making sure that binary is a string

oleg-andreyev:
it's possible to set NULL as binary and you'll get

```
Could not open connection: invalid argument: entry 0 of 'firstMatch' is invalid
from invalid argument: cannot parse capability: goog:chromeOptions
from invalid argument: cannot parse binary
from invalid argument: must be a string
```

### DIFF
--- a/lib/Chrome/ChromeOptions.php
+++ b/lib/Chrome/ChromeOptions.php
@@ -59,7 +59,7 @@ class ChromeOptions
      */
     public function setBinary($path)
     {
-        $this->binary = $path;
+        $this->binary = (string) $path;
 
         return $this;
     }


andrewnicols:
What are you passing it that it will cast to a valid string value?

I can't help but feel that you should be passing it a string anyway.

I suspect that type-hints will be added in 2.0 which would cause this to error if not passed a string too.

oleg-andreyev:
@andrewnicols type-declaration will help for sure. I caught this issue while working on https://github.com/oleg-andreyev/MinkFacebookWebDriver

williamdes:
@oleg-andreyev will you rename your project ?
`MinkFacebookWebDriver` > `MinkPhpWebDriver`

andrewnicols:
I was planning to rename mine to php-webdriver/mink or similar and put it in the same namespace as this project. Mine is fully feature complete and I’m just waiting for the 1.8 release to finish up final checks and release.

oleg-andreyev:
@williamdes I can do it, it's totally not a problem. @andrewnicols since we're doing similar things, I think it's worth of collaborating, I've done something differently and added some workaround. Right now I'm fighting with Travis setup, but I had quite decent results
https://travis-ci.com/oleg-andreyev/MinkFacebookWebDriver/builds/144076274

andrewnicols:
Happy to collaborate @oleg-andreyev. I have taken some different steps too in my implementation.

My intent is to ultimately require webdriver 2.0 and only those supported PHP versions. I've killed use of Syn (as have you) because it was used in an incorrect fashion. I'm also planning to switch to strict type declarations, and type hinting everywhere.

I have Travis fully integrated, but I have also found that the upstream Mink tests are fundamentally incorrect when it comes to the Key unit tests. They make incorrect assumptions as to what the correct keyboard events should be. I am taking the purist route and doing what is in the W3C specification rather than appeasing the Mink tests.

Right now my work has stalled as I busy, and waiting for 1.8.0 to be released. Once that happens and I have a little more free time my aim was to update the Mink driver for 1.8.0, tidy up versioning, and discuss with @OndraM about renaming the project to `php-webdriver/MinkDriver` or similar), and finally releasing properly on Packagist.

It looks like we have both taken several similar decisions:
# removing Syn
# Not relying upon `clear()` to empty

However it looks like we have both made different decisions on things like blurring (I do not use Tab, I send a browser event to meet the Mink spec).

andrewnicols:
My general feeling here is that only a string should be passed into the driver. If you are stringy object, then cast it to a string first. In version 2.0 of this driver I can well see us being much more strict.

oleg-andreyev:
@andrewnicols I've actually forked Mink Test Suite because it has a quite impressive amount of tests for Mink Driver and because I've been fixing many tests to be W3C compatible
You can take a look https://github.com/minkphp/driver-testsuite/compare/master...oleg-andreyev:integration-branch

OndraM:
This issue is in every single setter method. This hidden typecasting would lead to hiding errors in code. You can solve it be using assert library like [beberlei/assert](https://github.com/beberlei/assert) and assert data type inside the setter method. But I think better would to just add typehints in 2.0 and solve this issue in the whole codebase.

So I think for now it would be better to close this PR and solve it in 2.0 - and not add this just in this particular case.

oleg-andreyev:
@OndraM I agree. Do you need help with 2.0?

PR #734, opened 2020-01-14
Change license to MIT

OndraM:
We'd like to switch current license from Apache-2.0 to MIT license - which is simpler and more-widely used by the open-source community. Switching to MIT is something Facebook also did with eg. React, Jest, Flow, and Immutable.js. Thanks to [Facebook CLA](https://code.facebook.com/cla/) agreement, which all previous contributors must have signed, Facebook has the right to change the license. 

This is probably the only time we can do this, because starting from now (where new contributions will be without Facebook CLA signed), we could only change license in the future with agreement of all individual contributors (what would make it almost impossible).

So I ask @JoelMarcey from Facebook's open source team if he can approve this change on behalf of Facebook. :pray:

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,201 +1,22 @@
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+MIT License
 
-   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+Copyright (c) 2004-2020 Facebook
+Copyright (c) 2020-present [open-source contributors](https://github.com/php-webdriver/php-webdriver/graphs/contributors)
 
-   1. Definitions.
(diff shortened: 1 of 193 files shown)


OndraM:
Thanks 👍.

And that's true - I don't think it's necessary to have the header with copyright info, the license.md file should be enough and it only adds more noise when in each file.

andrewnicols:
+1 to do this, but I wonder whether there is some merit in still having some form of CLA.

OndraM:
> I wonder whether there is some merit in still having some form of CLA.

Do you have something specific on mind? I guess just MIT license together with (updated) contributing guidelines and code of conduct should be enough. CLA are AFAIK used sometimes when there is company or foundation behind the open source project and want to be able eg. dual-license it etc.

OndraM:
Merging this now (so that we can go forward with other PRs), but we can continue the discussion here if needed :).

PR #738, opened 2020-01-22
Custom commands support (v2)

OndraM:
Alternative implementation of Custom Commands, based on #611 .

The changes were only internal refactoring, so the usage is exactly the same as in #611:

```php
$clipboard = $driver->executeCustomCommand(
    '/session/:sessionId/appium/device/get_clipboard/',
    'POST'
);
```


OndraM:
Added to wiki: https://github.com/php-webdriver/php-webdriver/wiki/Custom-commands

PR #747, opened 2020-02-05
Upload files when W3cCompliant

krazzer:
In RemoteWebElement::sendKeys, a file is not uploaded when isW3cCompliant is true, resulting in attaching a file to a file field always giving the error "file not found", because the local file path is passed, instead of the uploaded file path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,66 +36,137 @@ Then install the library:
 
 ## Getting started
 
-### Start Server
+### 1. Start server (aka. remote end)
 
-The required server is the `selenium-server-standalone-#.jar` file provided here: http://selenium-release.storage.googleapis.com/index.html
+To control a browser, you need to start a *remote end* (server), which will listen to the commands sent
+from this library and will execute them in the respective browser.
 
-Download and run the server by **replacing #** with the current server version. Keep in mind **you must have Java 8+ installed to run this command**.
+This could be Selenium standalone server, but for local development, you can send them directly to so-called "browser driver" like Chromedriver or Geckodriver.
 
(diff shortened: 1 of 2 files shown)


OndraM:
Unfortunately, W3C WebDriver protocol does not yet include feature to remote file upload: https://github.com/w3c/webdriver/issues/1355

So it for example break Geckodriver build, as the feature is not present here. However it will work in Chrome, as this non-W3C endpoint is implemented here.

If we want to implement some workaround to make it work at least in Chrome, it needs to be done in a bit more complex way (now you would not be able upload even local file in Firefox) - let me check the options.

OndraM:
Replaced by #751

OndraM:
Merged, thanks @krazzer!

kirilkirkov:
@krazzer @OndraM 

When using my selenium code with Docker for file uploads, I encounter an error stating “invalid argument: File not found” for some files, especially those larger than 100MB. 

The exact error message:
**invalid argument: File not found : /Users/user/Downloads/11r/asd.flac
(Session info: chrome=120.0.6099.71)**

This issue doesn’t occur consistently, and I suspect it’s related to file size. Any advice on resolving this problem?

        
`sudo docker run -d -p 4444:4444 -p 5900:5900 --shm-size=2g --restart=always selenium/standalone-chrome:latest`
                
 ```
$fileInput = $this->driver->findElement(WebDriverBy::id("up-field"));
$fileInput->setFileDetector(new LocalFileDetector());
$fileInput->sendKeys('....');
```

PR #748, opened 2020-02-05
Let ZipArchive handle the creation of the zip files

rusakov92:
Hello, I've prepared a fix for an issue related to the new version of the PHP zip module. It was discussed in an open issue on the repository codeception/codeception, [link](https://github.com/Codeception/Codeception/issues/5838) to the issue.

In short, if we run codeception tests with Libzip 1.6 ZipArchive will fail to open the temp file and return error code 19.

This PR will instead let ZipArchive directly create the file for us.

### DIFF
--- a/lib/Remote/RemoteWebElement.php
+++ b/lib/Remote/RemoteWebElement.php
@@ -499,7 +499,7 @@ class RemoteWebElement implements WebDriverElement, WebDriverLocatable
         }
 
         // Create a temporary file in the system temp directory.
-        $temp_zip = tempnam(sys_get_temp_dir(), 'WebDriverZip');
+        $temp_zip = sys_get_temp_dir() . DIRECTORY_SEPARATOR . uniqid('WebDriverZip', false);
         $zip = new ZipArchive();
         if (($errorCode = $zip->open($temp_zip, ZipArchive::CREATE)) !== true) {
             throw new WebDriverException(sprintf('Error creating zip archive: %s', $errorCode));
--- a/lib/Remote/RemoteWebElement.php
+++ b/lib/Remote/RemoteWebElement.php
@@ -498,26 +498,38 @@ class RemoteWebElement implements WebDriverElement, WebDriverLocatable
             throw new WebDriverException('You may only upload files: ' . $local_file);
         }
 
+        $temp_zip_path = $this->createTemporaryZipArchive($local_file);
+
+        $remote_path = $this->executor->execute(
+            DriverCommand::UPLOAD_FILE,
+            ['file' => base64_encode(file_get_contents($temp_zip_path))]
+        );
+
+        unlink($temp_zip_path);
+
+        return $remote_path;
+    }
+
+    /**
+     * @param string $fileToZip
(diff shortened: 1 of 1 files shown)


OndraM:
Hi, first of all, please do not import consts and functions - it is not part of our codestyle.

And how does the proposed change any behavior?

For example, the previous code:
```
$temp_zip = tempnam(sys_get_temp_dir(), 'WebDriverZip');
```
does generate on my system something like `/tmp/WebDriverZip0L4oPp`.

While the new code does almost the same:
```
$temp_zip = sys_get_temp_dir() . DIRECTORY_SEPARATOR . uniqid('WebDriverZip', false);
```
`/tmp/WebDriverZip5e3bddac0101c` etc.

How is this supposed to fix the issue?

williamdes:
I think https://www.php.net/manual/en/function.tempnam.php can have strange behaviors in some cases and maybe not return a value

TLDR the original issue

OndraM:
Well I read the original issue, but I'm still not sure what is going on here - and I'd like to understand how how does this fix the issue...

Is the buggy `tempnam()` behavior described somewhere? When does it return null or anything which ZipArchive cannot handle?

rusakov92:
Hello, the issue comes from libzip, we recently updated it to version 1.6.0 and in their release notes mentioned that it won't accept empty files anymore, (1.6.0 notes)[https://github.com/nih-at/libzip/releases/tag/rel-1-6-0].

This affected other parts of our code base and the fix in this PR should work with all releases of libzip library.

OndraM:
Oh I see! The difference is `tempnam()` not only returns file name, it also have sideffects as it do create the (empty) file.

Could you @rusakov92 please update to codestyle, I will get this merged then.

rusakov92:
Thank you, once I have access to my PC I'll update the code style.

lib/Remote/RemoteWebElement.php
```diff
@@ -498,26 +498,38 @@ protected function upload($local_file)
...
+    /**
+     * @param string $fileToZip
+     * @return string
+     */
+    protected function createTemporaryZipArchive($fileToZip)
+    {
         // Create a temporary file in the system temp directory.
-        $temp_zip = tempnam(sys_get_temp_dir(), 'WebDriverZip');
+        // Intentionally do not use `tempnam()`, as it creates empty file which zip extension may not handle.
+        $tempZipPath = sys_get_temp_dir() . '/' . uniqid('WebDriverZip', false);
```
williamdes: why not using `DIRECTORY_SEPARATOR` ?

rusakov92: Looks like this was changed by @OndraM, it should be the DIRECTORY_SEPARATOR constant used here because it will break on systems running windows.

OndraM: Yes, I changed it intentionally, as  DIRECTORY_SEPARATOR is not needed, as "/" works on both unix and windows systems.

https://dev.to/c33s/always-use--as-directory-seperator-in-php-43l7

PR #752, opened 2020-02-06
Add few missing unit tests

OndraM:


williamdes:
@OndraM could you setup codecov or add a coverage badge?

OndraM:
We already use Coveralls, which is similar to codecov. However the coverage
is just about 50 % (even though it is increasing progressively), so I don't
think this is something to show off 😁.

Dne čt 6. 2. 2020 13:00 uživatel William Desportes <notifications@github.com>
napsal:

Issue #757, opened 2020-02-11
dragAndDropBy uses absolute position instead of offset in 1.8.0


wgevaert:
### What are you trying to achieve? (Expected behavior)
I am using php-webdriver for unit-testing, in which one test involves dragging and dropping a specific element. In 1.7.1, this was written as
```php
        $actions = new WebDriverActions($this->driver);
        $actions->dragAndDropBy(
            $this->element,
            0,
            -50
        )->perform();
```
and it worked perfectly.

### What do you get instead? (Actual behavior)
After the auto-update of composer to 1.8.0, the test threw a `MoveTargetOutOfBoundsException`. 

Some debugging attempts did not solve the problem. In 1.7.1, the `HttpCommandExecutor::execute` got the following inputs:
```json
$raw => {"method":"POST","url":"\/session\/:sessionId\/moveto"}
$http_method => "POST",
$url => "\/session\/3fcbbbc0446b4c1bbb7cae8e0424c3a5\/moveto",
$params => {"xoffset":0,"yoffset":50},
```
In 1.8.0 there seemed to be at least 2 relevant executed commands:
1: 
```json
$http_options => {"url":"\/session\/:sessionId\/actions","method":"POST"},
$http_mehtod => "POST"
$url => "\/session\/c38e68185246f7f29fe206c62d2a6c19\/actions"
$params => {"actions":[{"type":"pointer","id":"mouse","parameters":{"pointerType":"mouse"},"actions":[{"type":"pointerMove","duration":0,"x":0,"y":0,"origin":{"element-6066-11e4-a52e-4f735466cecf":"b72b8ae5-b44c-44be-8b27-d7fca391e562"}},{"type":"pointerDown","duration":0,"button":0}]}]}
```
2: 
```json
$http_options => {"url":"\/session\/:sessionId\/actions","method":"POST"}
$http_method => "POST"
$url => "\/session\/c38e68185246f7f29fe206c62d2a6c19\/actions"
$params => {"actions":[{"type":"pointer","id":"mouse","parameters":{"pointerType":"mouse"},"actions":[{"type":"pointerMove","duration":0,"x":0,"y":-50}]}]}]
```

The second execute gives the error. Maybe the problem is solved when this is also given an origin? I'm not sure.

### How could the issue be reproduced? (Steps to reproduce)

[...]

PR #759, opened 2020-02-11
Fix sendKeys not accepting array as input

### LINKED ISSUE #758: sendKeys on RemoteWebElement no longer accepts an array
I'm one of the maintainers of Dusk and we got this issue report just in: https://github.com/laravel/dusk/issues/732 after people started updating to the new 1.8.0 version of php-webdriver. It seems that the changes in https://github.com/php-webdriver/php-webdriver/pull/560 are causing some problems and changing the behavior of some methods. 

For example, we were sending an array of keys to the `sendKeys` but it doesn't accept that anymore when `$this->isW3cCompliant` is set to `true` in `RemoteWebElement`. 

From what I can tell this is set to true automatically based on the driver. The changelog for 1.8.0 notes:

> The protocol will be used automatically when remote end (like Geckodriver, newer Chromedriver etc.) supports it.

This seems unwanted because it changes the behavior (as noted above) on a minor release and thus causes breakages in upstream implementations (like Dusk). 

I'd suggest the automatic enabling of the protocol is reverted and only applied when a new major version is released.

That or the breaking change in this particular case is fixed when the `isW3cCompliant` is enabled. But that would be more work I guess and I don't know where there could be more breaking changes. 

Instead of allowing us to pass in an array like before we now get an array to string conversion exception.

I don't know the exact code to reproduce with only the php-webdriver but I guess that if you just use the chromedriver and were attempting to pass in an array to `sendKeys` on a `RemoteWebElement` instance before it'll now be broken. There's steps to reproduce the issue with dusk on this issue report: https://github.com/laravel/dusk/issues/732

### Details

* Php-webdriver version: 1.8.0
* PHP version: 7.4.2
* Selenium server version: unknown
* Operating system: unknown
* Browser used + version: unknown

I don't know about the last three versions above. I'll try to ask the person reporting the issue.

OndraM:
Fixes #758

### DIFF
--- a/lib/WebDriverKeys.php
+++ b/lib/WebDriverKeys.php
@@ -90,11 +90,13 @@ class WebDriverKeys
     const COMMAND = self::META;
 
     /**
-     * Encode input of `sendKeys()`.
+     * Encode input of `sendKeys()` to appropriate format according to protocol.
+     *
      * @param string|array|int|float $keys
-     * @return array
+     * @param bool $isW3cCompliant
+     * @return array|string
      */
-    public static function encode($keys)
+    public static function encode($keys, $isW3cCompliant = false)
     {
         if (is_numeric($keys)) {
             $keys = (string) $keys;
@@ -105,7 +107,11 @@ class WebDriverKeys
         }
 
         if (!is_array($keys)) {
-            return [];
+            if (!$isW3cCompliant) {
+                return [];
+            }
+
+            return '';
         }
 
         $encoded = [];
@@ -117,6 +123,10 @@ class WebDriverKeys
             $encoded[] = (string) $key;
         }
 
-        return $encoded;
+        if (!$isW3cCompliant) {
+            return $encoded;
+        }
+
+        return implode('', $encoded);
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 This project versioning adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+### Fixed
+- Accept array as possible input to `sendKeys()` method. (Unintentional BC break in 1.8.0.)
 
 ## 1.8.0 - 2020-02-10
 ### Added
--- a/lib/Remote/RemoteWebElement.php
+++ b/lib/Remote/RemoteWebElement.php
@@ -330,7 +330,7 @@ class RemoteWebElement implements WebDriverElement, WebDriverLocatable
         if ($local_file === null) {
             if ($this->isW3cCompliant) {
                 $params = [
-                    'text' => (string) $value,
+                    'text' => WebDriverKeys::encode($value, true),
                     ':id' => $this->id,
                 ];
             } else {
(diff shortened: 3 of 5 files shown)


lib/Remote/RemoteKeyboard.php
```diff
@@ -37,7 +37,7 @@ public function sendKeys($keys)
     {
         if ($this->isW3cCompliant) {
             $activeElement = $this->driver->switchTo()->activeElement();
-            $activeElement->sendKeys($keys);
+            $activeElement->sendKeys(WebDriverKeys::encode($keys, true));
```
andrewnicols: `RemoteWebElement::sendKeys()` is also encoding the values which can lead to a double-encoded value. Strictlly speaking, this doesn't matter because the purpose of `encode` function is largely that of normalisation. The encoded value can be safely re-encoded without any ill effect.

That said, if we were to make changes to `encode` in the future such that it can no longer be called on an already-encoded value, then `sendKeys` would be sending a double-encoded value and this may be difficult to detect.

I'd suggest that we either drop the `encode()` call here entirely (and in the OSS variant too if possible), or we add relevant unit tests to ensure that a value such as the following does not break:

```
$this->driver->getKeyboard()->sendKeys([
    [WebDriverKeys::SHIFT . 'h'],
    'elp',
]);
```

lib/WebDriverKeys.php
```diff
@@ -90,11 +90,13 @@ class WebDriverKeys
...
-     * Encode input of `sendKeys()`.
+     * Encode input of `sendKeys()` to appropriate format according to protocol.
+     *
      * @param string|array|int|float $keys
-     * @return array
+     * @param bool $isW3cCompliant
+     * @return array|string
      */
-    public static function encode($keys)
+    public static function encode($keys, $isW3cCompliant = false)
```
andrewnicols: For future, we should really rename this to normalise since that it what is happening. No values are encoded.

lib/Remote/RemoteKeyboard.php
OndraM: Nice catch, the call is really redundant here and should not be there.

The encode call needs to stay in the OSS version though, because it is not send via `sendKeys()` and it needs to be converted to array of chars (which is the way JsonWire protocol accepts text).

lordjancso:
@OndraM Thank you for the fix! Will you release this patch soon?

OndraM:
@lordjancso I'm also currently taking care of #757 which was introduced in 1.8.0 - so I'd like to release the fix for both of the issues at the same time.

lordjancso:
@OndraM Okay, thank you for the info.
Do you have any ETA of the release or should I lock to 1.7.1? My tests are completely broken with 1.8.0.

OndraM:
Sorry, we generally do not provide any ETAs. But if #757 builds goes well (there is already a pull request with the fix #764), is could be soon.

However you can already add dependency to your composer.json require section: 
```json
"php-webdriver/webdriver": "dev-master as 1.8.1"
```

and start using it right now.

PR #768, opened 2020-02-17
Disable the Devtools JSONView

chinleung:
By default, Firefox's JSONView is enabled, which causes issue when trying to parse a JSON response.

For instance, a JSON like this:

```json
{
    "id": 1,
    "className": "App\\User"
}
```

Would be rendered with lot of HTML in the source like the screenshot below:

image

To fix the issue, we currently have to disable the preference manually:

```php
$caps = DesiredCapabilities::firefox();

$profile = new FirefoxProfile;
$profile->setPreference('devtools.jsonview.enabled', false);
$profile->setPreference(FirefoxPreferences::READER_PARSE_ON_LOAD_ENABLED, false);

$caps->setCapability(FirefoxDriver::PROFILE, $profile);
```

### DIFF
--- a/lib/Firefox/FirefoxPreferences.php
+++ b/lib/Firefox/FirefoxPreferences.php
@@ -16,6 +16,8 @@ class FirefoxPreferences
     const READER_PARSE_ON_LOAD_ENABLED = 'reader.parse-on-load.enabled';
     /** @var string Browser homepage */
     const BROWSER_STARTUP_HOMEPAGE = 'browser.startup.homepage';
+    /** @var string Should the Devtools JSON view be enabled? */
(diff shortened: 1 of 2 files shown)


OndraM:
I'm only thinking whether this should be part of php-webdriver.

But I guess it is improbable anyone would actually need the JsonView UI in the browser when runnin in E2E tests, so I'd say this is good to go.

williamdes:
> I'm only thinking whether this should be part of php-webdriver, because it influences default browser behavior.
> 
> But I guess it is improbable anyone would actually need the JsonView UI in the browser when running in E2E tests, so I'd say this is good to go.

I agree it looks good to go
I do not know at what scale it influences the browser behavior

PR #770, opened 2020-02-22
Fix Firefox support on Mac OS Catalina

dunglas:
Firefox support is currently not working on Mac OS Catalina because of https://bugs.php.net/bug.php?id=79296. This PR introduces a workaround to make it working.


OndraM:
@dunglas I updated your branch, can you test it works now?

PR #772, opened 2020-02-24
Fix creating driver from session ID

### LINKED ISSUE #771: SendKeys receive error "invalid argument: 'text' must be a string" when reusing the chrome session
Able to use sendKeys() after reuse existing active chrome session for automation. 

Received "Facebook\WebDriver\Exception\InvalidArgumentException  : invalid argument: 'text' must be a string" error when trying to send text to input using sendKeys()

here are the sample code to reproduce the issue
```
        // create new session
        $host = 'http://localhost:9515';
        $driver = RemoteWebDriver::create(
                 $host, 
                 DesiredCapabilities::chrome(),
                 5000, 
                 60000
         );
        $driver->manage()->window()->maximize();
        $session_id = $driver->getSessionID();
        $driver->get('http://www.google.com');
        $driver->wait(5)->until(
             WebDriverExpectedCondition::visibilityOfElementLocated(
                    WebDriverBy::xpath('//input[@name="q"]')
             )
         )->sendKeys('Hi');
        sleep(1);
        
        // recreate the driver using previous session id
        $driver = RemoteWebDriver::createBySessionID($session_id, $host);
        $driver->get('http://www.google.com');
        $driver->wait(5)->until(
               WebDriverExpectedCondition::visibilityOfElementLocated(
                      WebDriverBy::xpath('//input[@name="q"]')
                )
         )->sendKeys('Error');
```

### Details
if the driver is freshly created from RemoteWebDriver::create(), it does not have any issue. However, when reuse previous chrome session by RemoteWebDriver::createBySessionID(), it run into error.

* Php-webdriver version: 1.8.1
* PHP version: 7.3.9
* Selenium server version: chromedriver_mac64 v80.0.3987.106
* Operating system: macOS 10.15.3
* Browser used + version: Google Chrome v80.0.3987.116

OndraM:
Fixes #771.

When using `createBySessionID()`, I didn't find simple and reliable way how to programatically determine whether the existing browser instance is in W3C/JsonWire mode, so it must be told via fifth parameter.

Two changes happend here:
- The method incorrectly read fourth parameter instead of fifth, making it impossible to override the default (JsonWire) mode
- The mode is now by default W3C when using  `createBySessionID()` - its because W3C mode is also enabled by default in `create()` method, so I think we can assume most people will now use the W3C mode. And if they do not (or they don't know or use both modes on different browsers, like we do in functional tests), they can save value of `isW3cCompliant()` and provide it via fifth parameter - [see tests](https://github.com/php-webdriver/php-webdriver/compare/master...OndraM:fix/create-from-session?expand=1#diff-ec1da9cd066dfececa8924c3e954cf36R95-R98) for reference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project versioning adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Reimplement element `equals()` method to be working in W3C mode.
 
+### Fixed
+- Properly read fifth parameter whether W3C compliant instance should be created when using `createBySessionID()`.
+
 ## 1.8.1 - 2020-02-17
 ### Fixed
 - Accept array as possible input to `sendKeys()` method. (Unintentional BC break in 1.8.0.)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This project versioning adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 ### Changed
 - Reimplement element `equals()` method to be working in W3C mode.
+- New instance of `RemoteWebDriver` created via `createBySessionID()` by default expects W3C mode. This could be disabled using fifth parameter of `createBySessionID()`.
 
 ### Fixed
 - Properly read fifth parameter whether W3C compliant instance should be created when using `createBySessionID()`.
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -10,4 +10,8 @@ parameters:
         - '#Call to an undefined method Facebook\\WebDriver\\WebDriverElement::equals\(\)#'
         - '#Unsafe usage of new static\(\)#'
 
+        # Parameter is intentionally not part of signature to not break BC
+        - message: '#PHPDoc tag \@param references unknown parameter: \$isW3cCompliant#'
+          path: 'lib/Remote/RemoteWebDriver.php'
+
     inferPrivatePropertyTypeFromConstructor: true
--- a/tests/functional/RemoteWebDriverCreateTest.php
+++ b/tests/functional/RemoteWebDriverCreateTest.php
@@ -92,11 +92,15 @@ class RemoteWebDriverCreateTest extends WebDriverTestCase
 
         // Store session ID
         $sessionId = $originalDriver->getSessionID();
+        $isW3cCompliant = $originalDriver->isW3cCompliant();
 
         // Create new RemoteWebDriver instance based on the session ID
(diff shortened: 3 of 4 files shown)


phpstan.neon
```diff
@@ -10,4 +10,8 @@ parameters:
         - '#Call to an undefined method Facebook\\WebDriver\\WebDriverElement::equals\(\)#'
         - '#Unsafe usage of new static\(\)#'
 
+        # Parameter is intentionally not part of signature to not break BC
```
williamdes: TODO: v2 clean the error

tests/functional/RemoteWebDriverCreateTest.php
```diff
@@ -92,11 +92,15 @@ public function testShouldCreateInstanceFromExistingSessionId()
 
         // Store session ID
         $sessionId = $originalDriver->getSessionID();
+        $isW3cCompliant = $originalDriver->isW3cCompliant();
 
         // Create new RemoteWebDriver instance based on the session ID
-        $this->driver = RemoteWebDriver::createBySessionID($sessionId, $this->serverUrl);
+        $this->driver = RemoteWebDriver::createBySessionID($sessionId, $this->serverUrl, null, null, $isW3cCompliant);
```
williamdes: from what I understand we should have 2 tests for the 2  number of possible arguments

phpstan.neon
OndraM: This is checked implicitly - once the root issue is resolved (param is part of signature), phpstan will trigger an error if the ignored error is no longer matched.

tests/functional/RemoteWebDriverCreateTest.php
OndraM: This is ensured by having test builds in both JsonWire and W3C mode, isn't it?

williamdes: Yes right ;)

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This project versioning adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 ### Changed
 - Reimplement element `equals()` method to be working in W3C mode.
+- New instance of `RemoteWebDriver` created via `createBySessionID()` by default expects W3C mode. This could be disabled using fifth parameter of `createBySessionID()`.
+
+### Fixed
+- Properly ready fifth parameter whether W3C compliant instance should be create when using `createBySessionID()`.
```
OndraM:
```suggestion
- Properly read fifth parameter whether W3C compliant instance should be created when using `createBySessionID()`.
```

PR #776, opened 2020-03-19
Clean up some files from dist version

williamdes:
image


williamdes:
See https://github.com/php-webdriver/php-webdriver/issues/783#issuecomment-618853997 how our files end up in each computer

PR #777, opened 2020-03-22
Add `all` and `any` expected conditions

barbaris-in:
Along with an `not` condition would be great to have `and` and `or` conditions. I called them `all` and `any` respectively

### DIFF
--- a/lib/WebDriverExpectedCondition.php
+++ b/lib/WebDriverExpectedCondition.php
@@ -6,6 +6,7 @@ use Facebook\WebDriver\Exception\NoAlertOpenException;
 use Facebook\WebDriver\Exception\NoSuchElementException;
 use Facebook\WebDriver\Exception\NoSuchFrameException;
 use Facebook\WebDriver\Exception\StaleElementReferenceException;
+use InvalidArgumentException;
 
 /**
  * Canned ExpectedConditions which are generally useful within webdriver tests.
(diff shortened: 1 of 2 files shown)


lib/WebDriverExpectedCondition.php
```diff
@@ -589,28 +589,31 @@ function (WebDriver $driver) use ($condition) {
      * An expectation with the logical AND condition of the given conditions.
      *
      * @param WebDriverExpectedCondition[] $conditions Array of the conditions to be satisfied.
-     * @return static Are all conditions satisfied or not.
+     * @return static Array of the getApply() values of all satisfied conditions or FALSE if some condition failed.
```
williamdes:
```suggestion
     * @return static|false Array of the getApply() values of all satisfied conditions or FALSE if some condition failed.
```

lib/WebDriverExpectedCondition.php
```diff
@@ -583,4 +584,64 @@ function (WebDriver $driver) use ($condition) {
...
+                return $results;
+            }
+        );
+    }
+
+    /**
+     * An expectation with the logical OR condition of the given conditions.
+     *
+     * @param WebDriverExpectedCondition[] $conditions Array of the conditions one of them to be satisfied.
+     * @return static Condition returns the return value of the getApply() of the first satisfied condition.
```
williamdes:
```suggestion
     * @return static|false Condition returns the return value of the getApply() of the first satisfied condition.
```

lib/WebDriverExpectedCondition.php
barbaris-in: To be precise these methods always return `static` and the comments are related to `apply` return values

barbaris-in:
Should I add some functional tests?

lib/WebDriverExpectedCondition.php
```diff
@@ -583,4 +584,65 @@ function (WebDriver $driver) use ($condition) {
             }
         );
     }
+
+    /**
+     * An expectation with the logical AND condition of the given conditions.
+     *
+     * @param WebDriverExpectedCondition[] $conditions Array of the conditions to be satisfied.
+     * @return static|false Array of the getApply() values of all satisfied conditions or FALSE if some condition
```
williamdes:
```suggestion
     * @return static array of the getApply() values of all satisfied conditions or FALSE if some condition
```
I am sorry, I misunderstood the code
There is no return false for the function but only a false value in the object
Could you make it clearer to understand ?

lib/WebDriverExpectedCondition.php
```diff
@@ -583,4 +584,65 @@ function (WebDriver $driver) use ($condition) {
...
+    {
+        return new static(
+            function (WebDriver $driver) use ($conditions) {
+                $results = [];
+                $allExist = true;
+
+                foreach ($conditions as $condition) {
+                    if (!$condition instanceof self) {
+                        throw new InvalidArgumentException(
+                            '$condition must be instance of WebDriverExpectedCondition class'
```
williamdes:
```suggestion
                            $condition . ' must be instance of WebDriverExpectedCondition class'
```
This might work, you are not used double quotes so the value would not be added in the string

lib/WebDriverExpectedCondition.php
```diff
@@ -583,4 +584,65 @@ function (WebDriver $driver) use ($condition) {
...
+     * @return static|false Condition returns the return value of the getApply() of the first satisfied condition.
+     */
+    public static function any($conditions)
+    {
+        return new static(
+            function (WebDriver $driver) use ($conditions) {
+                foreach ($conditions as $condition) {
+                    if (!$condition instanceof self) {
+                        throw new InvalidArgumentException(
+                            '$condition must be instance of WebDriverExpectedCondition class'
```
williamdes: same

williamdes:
> Should I add some functional tests?

@OndraM what do you recommend ?

lib/WebDriverExpectedCondition.php
barbaris-in: This won't work. `$condition` is an object of `WebDriverExpectedCondition` class which doesn't have `__toString()` method

PR #782, opened 2020-04-23
Allow to specify additional CURL options

mariopenterman:


lib/Remote/HttpCommandExecutor.php
```diff
@@ -210,6 +210,14 @@ public function __construct($url, $http_proxy = null, $http_proxy_port = null)
         $this->setRequestTimeout(30000);
         $this->setConnectionTimeout(30000);
     }
+    
+    /**
+     * Allows to specify the curl options
+     * @param $options
```
williamdes, line 216:
```suggestion
     * @param array $options
```
https://www.php.net/manual/en/function.curl-setopt-array.php

lib/Remote/HttpCommandExecutor.php
```diff
@@ -210,6 +210,14 @@ public function __construct($url, $http_proxy = null, $http_proxy_port = null)
...
+     */
+    public function setCurlOptions($options){
```
williamdes, line 218:
```suggestion
    public function setCurlOptions(array $options)
    {
```

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "php-webdriver/webdriver",
+  "name": "mariopenterman/php-webdriver",
```
williamdes, line 2: Sorry but please remove this from your PR

Issue #789, opened 2020-05-08
RemoteWebDriver::create() without $capabilities argument throws Exception


drjayvee:
```php
RemoteWebDriver::create($url);
```
throws
```
Facebook\WebDriver\Exception\InvalidArgumentException: invalid argument: entry 0 of 'firstMatch' must be a JSON object

~/repo/vendor/php-webdriver/webdriver/lib/Exception/WebDriverException.php:99
~/repo/vendor/php-webdriver/webdriver/lib/Remote/HttpCommandExecutor.php:370
~/repo/vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php:136
~/repo/tehst.php
```
even though the argument is optional.

* Php-webdriver version: 1.8.2
* PHP version: 7.3

PR #791, opened 2020-05-09
Fix RemoteWebDriver cannot be instantiated without second parameter

OndraM:
See #789

### DIFF
--- a/lib/Chrome/ChromeDriver.php
+++ b/lib/Chrome/ChromeDriver.php
@@ -35,7 +35,7 @@ class ChromeDriver extends RemoteWebDriver
             null,
             DriverCommand::NEW_SESSION,
             [
-                'desiredCapabilities' => $desired_capabilities->toArray(),
+                'desiredCapabilities' => (object) $desired_capabilities->toArray(),
             ]
         );
         $response = $this->executor->execute($command);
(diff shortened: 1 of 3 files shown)


OndraM:
It seems this won't be as easy :upside_down_face: .

For example, when using Selenium standalone server, it is IMHO expected to fail - as it obviously doesn't know which browser to start.
```
Facebook\WebDriver\Exception\SessionNotCreatedException: Unable to create session from {
  "desiredCapabilities": {
  },
  "capabilities": {
    "firstMatch": [
      {}
    ]
  }
}
```

andrewnicols:
Maybe we just make it a required param from now on?
API change, so 1.9 target?

OndraM:
> Maybe we just make it a required param from now on?
> API change, so 1.9 target?

API change would mean BC break... However, I think we should keep this allowed - eg. when you run your tests against local Chromedriver, you don't need to specify browser. So I think is it valid to send capabilities empty.

I updated the tests: they are skipped when using Selenium server (or Saucelabs). However they pass when run against specific browser driver.

williamdes:
> So I think is it valid to send capabilities empty.

Is it possible to define default capabilities on a selenium hub, and so sending empty ones would void the config ?

OndraM:
> > So I think is it valid to send capabilities empty.
> 
> Is it possible to define default capabilities on a selenium hub, and so sending empty ones would void the config ?

I don't think so... 🤔 IMHO you must always tell Selenium Hub which in browser you want to create the session.

Issue #790, opened 2020-05-08
alertIsPresent broken in W3C mode


sjerdo:
### What are you trying to achieve? (Expected behavior)
Wait for a popup to open

### What do you get instead? (Actual behavior)
NoSuchAlertException

### How could the issue be reproduced? (Steps to reproduce)
```php
                $webdriver->wait(5, 200)->until(
                    WebDriverExpectedCondition::alertIsPresent()
                );
```

### Details

* Php-webdriver version:  1.8.2
* PHP version: 7.3.12
* Selenium server version: 
* Operating system: 
* Browser used + version: ChromeDriver (v81.0.4044.138)

sjerdo:
alertIsPresent catches the wrong exception since this change:
https://github.com/php-webdriver/php-webdriver/commit/574684f29b8d513ae213c4567f2f71da92734db7

OndraM:
Hi @sjerdo and thanks for the report!

I can confirm this behavior for W3C protocol, the exceptions are indeed swapped. I'll prepare a fix and improve tests.

OndraM:
The fix is prepared in #792. :rocket:

OndraM:
The fix has been merged to master and will be part of next php-webdriver release.

Thanks @sjerdo for the report!

PR #792, opened 2020-05-09
Fix alertIsPresent in W3C mode

OndraM:
The catched exception was swapped. We can rely on `NoSuchAlertException` (w3c exception), because it catches also `NoAlertOpenException` (oss mode exception), as it is inherited from `NoSuchAlertException`.


williamdes:
#790

PR #799, opened 2020-06-18
Add new window command

andrewnicols:
The “New window” command (https://www.w3.org/TR/webdriver/#new-window) is not currently supported.

I’m not happy with this PR in its current form, mainly because I can’t work out the best return type, and I won’t write tests until then.

The remote endpoint returns an object containing:
- `handle`; and 
- `type` (tab, window)

In similar cases we return new objects like `WebDriverDimension` which contain the relevant data and but I’m not sure that there’s any point in returning a new `WebDriverWindow` element of any kind, because the Window performs the same as Frames - i.e. it sets the current top-level context.

I feel the best solution here is to return just the data - either just the new handle as a string, or both the handle and the type. I’m not sure if it’s worth creating a new WebDriverHandle for this.

Happy to take guidance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This project versioning adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 ### Added
 - Support of SameSite cookie property.
+- Command `RemoteWebDriver::newWindow()` for W3C mode to open new top-level browsing context (aka window).
 
 ## 1.8.3 - 2020-10-06
 ### Fixed
--- a/lib/Remote/DriverCommand.php
+++ b/lib/Remote/DriverCommand.php
@@ -140,6 +140,7 @@ class DriverCommand
     const ACTIONS = 'actions';
     const GET_ELEMENT_PROPERTY = 'getElementProperty';
     const GET_NAMED_COOKIE = 'getNamedCookie';
+    const NEW_WINDOW = 'newWindow';
     const TAKE_ELEMENT_SCREENSHOT = 'takeElementScreenshot';
     const MINIMIZE_WINDOW = 'minimizeWindow';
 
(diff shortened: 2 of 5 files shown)


OndraM:
Maybe have a look how this is handled by java bindings?

Also, how is it that we missed this endpoint from implementing previously :thinking:? Maybe we should cross-check the whole W3C spec (both level 1 and level 2) and write down all what is missing from full implementation of those two WebDriver spec level. I guess there is more of such missing features...

andrewnicols:
I did actually run through the entire spec a while back and compare against the spec - we fixed most of them in #683 but apparently I missed some.

That command was added in https://github.com/w3c/webdriver/commit/d1769cafc80132788c0cf1231170c35c8a3f39e7 on October 2018 (https://github.com/w3c/webdriver/pull/1324).

Perhaps I was working against an old draft of the specification.

Good idea re looking at the Java client.

andrewnicols:
Okay, so the Selenium client:
1. calls window/new
1. calls switchTo($handle)
1. return $this;

https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java#L925-L928

andrewnicols:
Okay,

I’ve pushed a new commit which adds the new window command. The only thing I’m not entirely happy about is the naming of the new function. I’ve currently gone for `window` but I think that’s too confusing. Perhaps `newWindow`?

andrewnicols:
I'm not quite sure how we're seeing a decrease in coverage. I think it's because of the W3C-only nature of the patch. Non-W3C clients will see a decrease in coverage as they are not able to run the lines (correctly).

OndraM:
Rebased to current main branch.

OndraM (approved):

OndraM:
Finally merged. another part of W3C implemented, thanks @andrewnicols! :1st_place_medal:

PR #800, opened 2020-06-18
Rename default branch to main

OndraM:
We plan to change default branch from "master" to "main". Some reasons behind these changes were covered by @localheinz here: https://localheinz.com/blog/2020/06/12/black-lives-matter/ 

We aim to have welcoming environment in this project, and if this change from possibly offensive language could help, then lets do it! Furthermore, the term `main` is semantically more explicit for what its used.

Also this should have minimal downsides: to notable one is for people relying on `dev-master` in composer.json, which will end with broken build. Using `dev-master` is not a good practice (constrain should be on version range, not on branch), but the only thing needed to fix this issue is in composer.json:

```diff
  "require": {
-    "php-webdriver/webdriver": "dev-master"
+    "php-webdriver/webdriver": "dev-main"
  },
```

As `main` branch is already pushed into repo, this could be used right now.

Once this PR is merged, I will change GitHub settings to have "main" as default branch, retarget active PRs and delete master branch few days afterwards (to have short transition period).

- [ ] Merge #800 
- [ ] Set default branch on github
- [ ] Retarget pull requests
- [ ] Remove master branch


OndraM:
The master branch has been removed.

PR #803, opened 2020-06-23
WebDriverSingleKeyAction will throw an exception if not a modifier

oleg-andreyev:
while it's not defined in specs directly, it makes sense to throw an exception from WebDriverSingleKeyAction if key is not a modifier

Ref.:
https://github.com/SeleniumHQ/selenium/blob/e561ca8672b199ca6931b46b32290ee7ca77cf3e/java/client/src/org/openqa/selenium/interactions/internal/SingleKeyAction.java#L49
https://github.com/SeleniumHQ/selenium/blob/00de1c6b652278181721f3f4e2373b6d0088795d/javascript/node/selenium-webdriver/lib/actions.js#L59

What do you think?

To Do:
- [x] Test

### DIFF
--- a/lib/Interactions/Internal/WebDriverKeyDownAction.php
+++ b/lib/Interactions/Internal/WebDriverKeyDownAction.php
@@ -2,9 +2,7 @@
 
 namespace Facebook\WebDriver\Interactions\Internal;
 
-use Facebook\WebDriver\WebDriverAction;
-
-class WebDriverKeyDownAction extends WebDriverSingleKeyAction implements WebDriverAction
+class WebDriverKeyDownAction extends WebDriverSingleKeyAction
 {
     public function perform()
     {
--- a/lib/Interactions/Internal/WebDriverSingleKeyAction.php
+++ b/lib/Interactions/Internal/WebDriverSingleKeyAction.php
@@ -5,13 +5,33 @@ namespace Facebook\WebDriver\Interactions\Internal;
 use Facebook\WebDriver\Internal\WebDriverLocatable;
 use Facebook\WebDriver\WebDriverAction;
 use Facebook\WebDriver\WebDriverKeyboard;
+use Facebook\WebDriver\WebDriverKeys;
 use Facebook\WebDriver\WebDriverMouse;
 
 abstract class WebDriverSingleKeyAction extends WebDriverKeysRelatedAction implements WebDriverAction
 {
-    /** @var string */
-    protected $key = '';
+    const MODIFIER_KEYS = [
+        WebDriverKeys::SHIFT,
+        WebDriverKeys::LEFT_SHIFT,
+        WebDriverKeys::RIGHT_SHIFT,
+        WebDriverKeys::CONTROL,
+        WebDriverKeys::LEFT_CONTROL,
+        WebDriverKeys::RIGHT_CONTROL,
+        WebDriverKeys::ALT,
+        WebDriverKeys::LEFT_ALT,
+        WebDriverKeys::RIGHT_ALT,
+        WebDriverKeys::META,
+        WebDriverKeys::RIGHT_META,
+        WebDriverKeys::COMMAND,
+    ];
 
+    /** @var string */
+    protected $key;
+
+    /**
+     * @param string $key
+     * @todo Remove default $key value in next major version (BC)
+     */
     public function __construct(
         WebDriverKeyboard $keyboard,
         WebDriverMouse $mouse,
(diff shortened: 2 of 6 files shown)


lib/Interactions/Internal/WebDriverKeyDownAction.php
```diff
@@ -2,9 +2,7 @@
 
 namespace Facebook\WebDriver\Interactions\Internal;
 
-use Facebook\WebDriver\WebDriverAction;
-
-class WebDriverKeyDownAction extends WebDriverSingleKeyAction implements WebDriverAction
```
andrewnicols, line 7: Why remove the Interface for these two files?

lib/Interactions/Internal/WebDriverSingleKeyAction.php
```diff
@@ -5,10 +5,17 @@
 use Facebook\WebDriver\Internal\WebDriverLocatable;
 use Facebook\WebDriver\WebDriverAction;
 use Facebook\WebDriver\WebDriverKeyboard;
+use Facebook\WebDriver\WebDriverKeys;
 use Facebook\WebDriver\WebDriverMouse;
 
 abstract class WebDriverSingleKeyAction extends WebDriverKeysRelatedAction implements WebDriverAction
 {
+    const MODIFIER_KEYS = [
+        WebDriverKeys::SHIFT, WebDriverKeys::CONTROL, WebDriverKeys::ALT, WebDriverKeys::META,
```
andrewnicols, line 14: This should include the right modifiers too.

lib/Interactions/Internal/WebDriverKeyDownAction.php
oleg-andreyev: WebDriverSingleKeyAction already implements WebDriverAction

lib/Interactions/Internal/WebDriverSingleKeyAction.php
oleg-andreyev: As of today, the left modifiers are just aliases to right modifies.
I've been thinking to create PR for that.

Ref.:
https://github.com/php-webdriver/php-webdriver/blob/e4347e2164d56c9ad20911f37d41919a0eaadf8a/lib/WebDriverKeys.php#L83-L90

andrewnicols:
Oh, I meant to say that this needs unit tests.

lib/Interactions/Internal/WebDriverKeyDownAction.php
andrewnicols: That should really be a separate pull request or, at the very least, a separate commit. It’s unrelated to the change in this PR.

lib/Interactions/Internal/WebDriverSingleKeyAction.php
```diff
@@ -5,10 +5,17 @@
...
+        WebDriverKeys::COMMAND, WebDriverKeys::LEFT_ALT, WebDriverKeys::LEFT_CONTROL,
+        WebDriverKeys::LEFT_SHIFT
```
OndraM, line 16:
```suggestion
        WebDriverKeys::SHIFT,
        WebDriverKeys::CONTROL,
        WebDriverKeys::ALT,
        WebDriverKeys::META,
        WebDriverKeys::COMMAND,
        WebDriverKeys::LEFT_ALT,
        WebDriverKeys::LEFT_CONTROL,
        WebDriverKeys::LEFT_SHIFT,
```

lib/Interactions/Internal/WebDriverSingleKeyAction.php
```diff
@@ -20,5 +27,9 @@ public function __construct(
     ) {
         parent::__construct($keyboard, $mouse, $location_provider);
         $this->key = $key;
+
+        if (!in_array($key, self::MODIFIER_KEYS, true)) {
```
OndraM, line 31: I'd put this assert beggining of the method, right after parent constructor call - this is usual place to put check for input conditions.

lib/Interactions/Internal/WebDriverSingleKeyAction.php
```diff
@@ -20,5 +27,9 @@ public function __construct(
...
+            throw new \InvalidArgumentException("Key Down / Up events only make sense for modifier keys.");
```
OndraM, line 32: I would slightly reword the exception to something like: 

```suggestion
            throw new \InvalidArgumentException('keyDown / keyUp actions can only be used for modifier keys');
```

oleg-andreyev:
> Oh, I meant to say that this needs unit tests.

added and updated.

OndraM (approved):
Merging this, thanks! Sorry it took so long.

PR #804, opened 2020-06-23
Window API will return Window Rect

oleg-andreyev:
based on specifications Window API, window size manipulation returns window rect, which could be useful for consumer

Ref.:
- https://www.w3.org/TR/webdriver/#get-window-rect
- https://www.w3.org/TR/webdriver/#set-window-rect
- https://www.w3.org/TR/webdriver/#maximize-window
- https://www.w3.org/TR/webdriver/#minimize-window
- https://www.w3.org/TR/webdriver/#fullscreen-window

Because it's an API change, it can only go into the main branch and major version release.

To Do:
- [ ] tests


williamdes (changes requested):
This PR adds BC breaks, maybe -> v2

oleg-andreyev:
Yes, I think this was discussed with @OndraM in PR comments

oleg-andreyev:
> Yes, I think this was discussed with @OndraM in PR comments 

To be exact @OndraM added milestone 2.0

williamdes:
> > Yes, I think this was discussed with @OndraM in PR comments
> 
> To be exact @OndraM added milestone 2.0

Sorry, I missed that

williamdes (approved):
💯

composer.json
```diff
@@ -28,7 +28,7 @@
     "php-coveralls/php-coveralls": "^2.4",
     "php-mock/php-mock-phpunit": "^1.1 || ^2.0",
     "php-parallel-lint/php-parallel-lint": "^1.2",
-    "phpunit/phpunit": "^5.7 || ^7 || ^8 || ^9",
+    "phpunit/phpunit": "^5.7,>=5.7.15 || ^7 || ^8 || ^9",
```
williamdes, line 31:
```suggestion
    "phpunit/phpunit": "^5.7,>=5.7.15 || ^7 || ^8 || ^9",
```
Could you explain why this is need please ?
it looks wrong because `^5.7` would include `>=5.7.15`

oleg-andreyev:
@OndraM please take a look

composer.json
oleg-andreyev: phpunit < 5.7.15 has a bug with yield.

uuf6429: Verrrrry late reply, you can simply use `^5.7.15`, which is perhaps less ambiguous than the current one.

PR #806, opened 2020-07-01
Update WebDriverActions.php dragAndDropBy

jukimv1986:
This fix was needed when interacting with BrowserStack Selenium servers.

### DIFF
--- a/lib/Interactions/WebDriverActions.php
+++ b/lib/Interactions/WebDriverActions.php
@@ -144,10 +144,10 @@ class WebDriverActions
             new WebDriverClickAndHoldAction($this->mouse, $source)
         );
         $this->action->addAction(
-            new WebDriverMoveToOffsetAction($this->mouse, null, $x_offset, $y_offset)
+            new WebDriverMoveToOffsetAction($this->mouse, $source, $x_offset, $y_offset)
         );
         $this->action->addAction(
-            new WebDriverButtonReleaseAction($this->mouse, null)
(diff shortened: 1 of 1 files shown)


williamdes:
Hi @jukimv1986 could you explain in details the before and after behavior ?
We use/used BrowserStack at phpMyAdmin so I am very interested by understanding this fix

andrewnicols:
This change does not make sense to me.

The function where you make these changes adds three actions:

```
new WebDriverClickAndHoldAction($this->mouse, $source);
new WebDriverMoveToOffsetAction($this->mouse, null, $x_offset, $y_offset);
new WebDriverButtonReleaseAction($this->mouse, null);
```

So essentially the source is clicked and the mouse buttonheld.
Then the mouse is moved by the x/y offset
And finally released.

The `null` option mean that the origin is set to `pointer` as per the W3C specification at https://www.w3.org/TR/webdriver/#pointer-actions

When pointer is used, the x and y offset are used for the move as per:

    Let x equal start x + x offset and y equal start y + y offset

So the proposed change in the call to {{WebDriverMoveToOffsetAction}} is wrong - that is _meant_ to be an offset and _not_ relative to a source object.

The final line changed is purely a release with no move and source makes no sense at all.

Can you please describe the bug that you are fixing.

andrewnicols:
Hi @jukimv1986,

Do you have any more information on this proposed change?

As I understand it, the current behaviour is correct but I ma be missing some part of the puzzle.

Thanks,

Andrew

PR #815, opened 2020-08-17
Add DriverServerDied exception

### LINKED ISSUE #595: Usefull Exception messages are discareded
In `\Facebook\WebDriver\Remote\Service\DriverService::isRunning` we have 
```
            if (!$this->service->isRunning()) {
                throw new WebDriverException('The driver server has died.');
            }
```

This blocks spoilers the meaningful error message e.g. `no such element: Unable to locate element: {"method":"id","selector":"[aria-label*=Log]"}`  (from `Facebook\WebDriver\Exception\NoSuchElementException` with a hardcoded error message `The driver server has died`. That make debugging hard.  Line 65 (`throw $e;`) is not reachable in any way when `this->service->isRunning()` is true.  I can creat MR to fix it. 

* Php-webdriver version:  1.6.0
* PHP version: 7.1
* ChromeDriver 
* Operating system: Win 10 
* Browser used:  Chrome

TomAdam:
This PR adds a new `DriverServerDiedException` with the previous error chained to improve debugging of 'Driver server has died.' exceptions. Note that line 15 of `DriverServerDiedException` is nasty but it is needed to avoid a BC break as `DriverCommandExecutor::execute()` currently throws a `WebDriverException`. 

Alternatively a named constructor could be used in `WebDriverException` to allow injection of the previous exception. It didn't feel like it fit in this class as it isn't a driver exception but rather a process control exception. Let me know if you would prefer this.

Fixes #595.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 This project versioning adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+### Changed
+- Throw `DriverServerDiedException` on driver server process terminating unexpectedly in `DriverCommandExecutor`, with triggering error attached to improve debugging of 'The driver server has died' errors. 
 
 ## 1.9.0 - 2020-11-19
 ### Added
--- a/lib/Exception/DriverServerDiedException.php
+++ b/lib/Exception/DriverServerDiedException.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Facebook\WebDriver\Exception;
+
+/**
+ * The driver server process is unexpectedly no longer available.
+ */
+class DriverServerDiedException extends WebDriverException
+{
+    public function __construct(\Exception $previous = null)
+    {
+        parent::__construct('The driver server has died.');
+        \Exception::__construct($this->getMessage(), 0, $previous);
+    }
+}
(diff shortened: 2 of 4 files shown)


TomAdam:
Slight drop in testing coverage due to the added code. It looks like it would be challenging to add this due to the call to `parent::execute()` in `Facebook\WebDriver\Remote\Service\DriverCommandExecutor`.

TomAdam:
We are currently running this change in production and it caught a few errors today. We still aren't much closer to what is happening but we do now have more information than "Driver server has died".

```
Facebook\WebDriver\Exception\DriverServerDiedException The driver server has died.

Caused by: Facebook\WebDriver\Exception\WebDriverCurlException Curl error thrown for http POST to /session/d8b23b8f4dc3c9383b80bcd366866aa9/url with params: {****}

Failed to connect to localhost port 9528: Connection refused 
```

TomAdam:
Is there anything I can do to help this along? No pressure, just want to lend a hand if there is something useful I can do.

williamdes:
> Is there anything I can do to help this along? No pressure, just want to lend a hand if there is something useful I can do.

Thank you :)
Maybe @OndraM has an idea ?

OndraM (approved):
Thanks @TomAdam! I added unit test for the class, otherwise LGTM, nice job.

Now waiting until CI build finishes :)

williamdes (approved):

OndraM:
Merged to `main` and will be part of 1.10.0 once released. Thanks @TomAdam for your help! :+1:

PR #820, opened 2020-09-18
Add support for same-site cookies

### LINKED ISSUE #705: Missing "SameSite" attribute in cookies
If I have a cookie from webdriver I expect to read the "SameSite" attribute of the cookie.

In https://github.com/facebook/php-webdriver/blob/community/lib/Cookie.php#L49 the "SameSite" attribute is not saved in the cookie.

### Details

* Php-webdriver version: 1.7.1
* PHP version: 7.3.11
* Operating system: CentOS 7
* Browser used + version: Chromium 77.0.3865.90
* Chromedriver used: ChromeDriver 77.0.3865.90

codegain:
Fixes #705 

I tested this in a VM on Ubuntu 16.04 with the `chromium-browser` and the `chromedriver` packages:

```
Chromium 85.0.4183.83
ChromeDriver 85.0.4183.83
```

It correctly sets the values `Lax` and `Strict` (and `null` if `sameSite` is not set on the cookie). I also think I didn't come up with a good description for `setSameSite`, maybe someone comes up with a better one :)

### DIFF
--- a/lib/Cookie.php
+++ b/lib/Cookie.php
@@ -58,6 +58,9 @@ class Cookie implements \ArrayAccess
         if (isset($cookieArray['httpOnly'])) {
             $cookie->setHttpOnly($cookieArray['httpOnly']);
         }
+        if (isset($cookieArray['sameSite'])) {
+            $cookie->setSameSite($cookieArray['sameSite']);
+        }
 
         return $cookie;
     }
@@ -170,6 +173,24 @@ class Cookie implements \ArrayAccess
     public function isHttpOnly()
     {
         return $this->offsetGet('httpOnly');
+    }
+
+    /**
+     * The cookie's same-site value.
+     *
+     * @param string $sameSite
+     */
+    public function setSameSite($sameSite)
+    {
+        $this->offsetSet('sameSite', $sameSite);
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getSameSite()
+    {
+        return $this->offsetGet('sameSite');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lib/Cookie.php
```diff
@@ -172,6 +175,24 @@ public function isHttpOnly()
         return $this->offsetGet('httpOnly');
     }
 
+    /**
+     * The cookie's same-site value.
+     *
+     * @param string $sameSite
```
williamdes, line 181: Maybe should propose null too to redet the value?

codegain: All setters don't specifically define that `null` could be set. For example, the setter for "setHttpOnly" only "allowes" boolean values, but "isHttpOnly" is defined as returning boolean and `null`. However I don't have a strong opinion about adding the additional `null` param type.

OndraM:
Hi @codegain , thanks for the PR! Could you also please add unit tests? (`tests/unit/CookieTest.php` needs to be extended)

I also rebased the branch, so that it contain latest changes from the `main` branch.

codegain:
@OndraM I just added the unit tests but I was not quite sure, if the `sameSite` should also be tested in the method `testShouldNotContainNullValues`. Could you maybe take a look?

OndraM:
> @OndraM if the `sameSite` should also be tested in the method `testShouldNotContainNullValues`.

I think so, the optional parameters should be omitted, so this should be part of the test as well.

codegain:
> I think so, the optional parameters should be omitted, so this should be part of the test as well.

@OndraM And now it is :)

OndraM:
I added few enhancements, see #824.

SameSite cookie value is validated to be only "Lax", "Strict" or "None".

Also added constants for this value:

```php
$cookie = new Cookie('cookieName', 'cookie value');
$cookie->setSameSite(Cookie::SAMESITE_LAX); // or Cookie::SAMESITE_STRICT, Cookie::SAMESITE_NONE
```

PR #824, opened 2020-10-06
Validate SameSite cookie value

OndraM:
SameSite cookie value is validated to be only "Lax", "Strict" or "None". Extends #820 .

Also added constants for this value:

```php
$cookie = new Cookie('cookieName', 'cookie value');
$cookie->setSameSite(Cookie::SAMESITE_LAX);
```

### DIFF
--- a/lib/Cookie.php
+++ b/lib/Cookie.php
@@ -13,6 +13,13 @@ use InvalidArgumentException;
  */
 class Cookie implements \ArrayAccess
 {
+    /** @var string */
+    const SAMESITE_LAX = 'Lax';
+    /** @var string */
+    const SAMESITE_STRICT = 'Strict';
+    /** @var string */
+    const SAMESITE_NONE = 'None';
+
     /** @var array */
     protected $cookie = [];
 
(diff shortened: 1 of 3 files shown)


andrewnicols (approved):
These changes look good to me.

I considered whether we should be more lax and allow `strict` or `STRICT` etc. and normalise the values, but on reflection this is pointless. People should just use the correct values.

tests/unit/CookieTest.php
```diff
@@ -155,17 +155,18 @@ public function testShouldBeCreatableFromAnArrayWithAllValues()
...
     }
 
     /**
      * @dataProvider provideInvalidCookie
+     * @param string $expectedMessage
      * @param string $name
      * @param string $value
      * @param string $domain
-     * @param string $expectedMessage
+     * @param null $sameSite
```
williamdes, line 167:
```suggestion
     * @param string|null $sameSite
```

OndraM: Fixed, thanks!

lib/Cookie.php
```diff
@@ -182,6 +189,12 @@ public function isHttpOnly()
      */
```
williamdes, line 189: We could add a `@phpstan-param Cookie::*`

williamdes (approved):

PR #556, opened 2018-03-27
Add support for sending Chrome DevTools commands

aboks:
ChromeDriver has two endpoints for calling the DevTools API. This PR adds support for calling these endpoints (sending DevTools commands) using php-webdriver.

PR #828, opened 2020-10-30
PHP 8 Support

driesvints:
Working on providing full PHP 8 support for php-webdriver. We'll need to for Laravel Dusk.

Currently using this PR to run the builds.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -52,7 +52,7 @@ return PhpCsFixer\Config::create()
         'ordered_class_elements' => true,
         'ordered_imports' => true,
         'php_unit_construct' => true,
-        'php_unit_dedicate_assert' => true,
+        'php_unit_dedicate_assert' => false,
         'php_unit_expectation' => ['target' => '5.6'],
         'php_unit_method_casing' => ['case' => 'camel_case'],
         'php_unit_mock' => true,
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: php
 dist: xenial
 
 php:
-  - '5.6'
-  - '7.0'
   - '7.1'
   - '7.2'
   - '7.3'
@@ -21,10 +19,25 @@ services:
 
 matrix:
   include:
+    - name: 'PHP nightly build'
+      php: 'nightly'
+      env: DEPENDENCIES="--ignore-platform-req=php"
+
+    - name: 'PHP 7.0'
+      php: '7.0'
+      env: RUN_PHPUNIT_PATCHES="true"
+
+    - name: 'PHP 5.6'
+      php: '5.6'
+      env: RUN_PHPUNIT_PATCHES="true"
+
     # Build with lowest possible dependencies on lowest possible PHP
     - name: 'Lowest dependencies build'
       php: '5.6'
-      env: DEPENDENCIES="--prefer-lowest"
+      env:
+        - DEPENDENCIES="--prefer-lowest"
+        - RUN_PHPUNIT_PATCHES="true"
+
 
     # Firefox inside Travis environment
     - name: 'Firefox 45 on Travis (OSS protocol); via legacy Selenium server'
@@ -80,6 +93,7 @@ install:
   - travis_retry composer update --no-interaction $DEPENDENCIES
 
 before_script:
+  - if [ "$RUN_PHPUNIT_PATCHES" = "true" ]; then ./scripts/apply-phpunit-patches.sh; fi
   - if [ "$BROWSER_NAME" = "chrome" ]; then
       mkdir chromedriver;
       CHROME_VERSION=$(google-chrome --product-version);
@@ -126,4 +140,4 @@ after_script:
   - if [ -f ./logs/chromedriver.log ]; then cat ./logs/chromedriver.log; fi
 
 after_success:
-  - travis_retry php vendor/bin/php-coveralls -v
+  - if [ "$TRAVIS_PHP_VERSION" != "nightly" ]; then travis_retry php vendor/bin/php-coveralls -v; fi
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
   "homepage": "https://github.com/php-webdriver/php-webdriver",
   "license": "MIT",
   "require": {
-    "php": "^5.6 || ~7.0",
+    "php": "^5.6 || ~7.0 || ^8.0",
     "ext-curl": "*",
     "ext-json": "*",
     "ext-zip": "*",
@@ -25,12 +25,10 @@
   "require-dev": {
     "friendsofphp/php-cs-fixer": "^2.0",
     "ondram/ci-detector": "^2.1 || ^3.5",
-    "php-coveralls/php-coveralls": "^2.0",
-    "php-mock/php-mock-phpunit": "^1.1",
+    "php-coveralls/php-coveralls": "^2.4",
+    "php-mock/php-mock-phpunit": "^1.1 || ^2.0",
     "php-parallel-lint/php-parallel-lint": "^1.2",
-    "phpunit/phpunit": "^5.7",
-    "sebastian/environment": "^1.3.4 || ^2.0 || ^3.0",
-    "sminnee/phpunit-mock-objects": "^3.4",
+    "phpunit/phpunit": "^5.7 || ^7 || ^8 || ^9",
     "squizlabs/php_codesniffer": "^3.5",
     "symfony/var-dumper": "^3.3 || ^4.0 || ^5.0"
   },
--- a/scripts/apply-phpunit-patches.sh
+++ b/scripts/apply-phpunit-patches.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+# All commands below must no fail
+set -e
+
+# Be in the root dir
+cd "$(dirname $0)/../"
+
+find tests/ -type f -print0 | xargs -0 sed -i 's/function setUpBeforeClass(): void/function setUpBeforeClass()/g';
+find tests/ -type f -print0 | xargs -0 sed -i 's/function setUp(): void/function setUp()/g';
+find tests/ -type f -print0 | xargs -0 sed -i 's/function tearDown(): void/function tearDown()/g';
+
+sed -i 's/endTest(\\PHPUnit\\Framework\\Test \$test, float \$time): void/endTest(\\PHPUnit_Framework_Test \$test, \$time)/g' tests/functional/ReportSauceLabsStatusListener.php;
+sed -i 's/: void/ /g' tests/functional/ReportSauceLabsStatusListener.php;
+# Drop the listener from the config file
+sed -i '/<listeners>/,+2d' phpunit.xml.dist;
+sed -i 's/function runBare(): void/function runBare()/g' tests/functional/WebDriverTestCase.php;
+
+# Return back to original dir
+cd - > /dev/null
--- a/tests/functional/WebDriverTestCase.php
+++ b/tests/functional/WebDriverTestCase.php
@@ -29,7 +29,7 @@ class WebDriverTestCase extends TestCase
     /** @var int */
     protected $requestTimeout = 60000;
 
-    protected function setUp()
+    protected function setUp(): void
     {
         $this->desiredCapabilities = new DesiredCapabilities();
 
@@ -78,7 +78,7 @@ class WebDriverTestCase extends TestCase
         }
     }
 
-    protected function tearDown()
+    protected function tearDown(): void
     {
         if ($this->driver instanceof RemoteWebDriver && $this->driver->getCommandExecutor()) {
             try {
@@ -126,7 +126,7 @@ class WebDriverTestCase extends TestCase
      * Rerun failed tests.
      * @TODO Replace with PHPUnit 7.3+ builtin functionality once upgraded to PHP 7.1+
      */
-    public function runBare()
+    public function runBare(): void
     {
         $e = null;
         $numberOfRetires = 3;
@@ -212,5 +212,21 @@ class WebDriverTestCase extends TestCase
                 $this->desiredCapabilities->setCapability('build', $build);
             }
         }
+    }
+
+    /**
+     * Uses assertStringContainsString when it is available or uses assertContains for old phpunit versions
+     * @param string $needle
+     * @param string $haystack
+     * @param string $message
+     */
+    protected function compatAssertStringContainsString($needle, $haystack, $message = '')
+    {
+        if (method_exists($this, 'assertStringContainsString')) {
+            parent::assertStringContainsString($needle, $haystack, $message);
+
+            return;
+        }
+        parent::assertContains($needle, $haystack, $message);
     }
 }
--- a/tests/unit/Interactions/Internal/WebDriverCoordinatesTest.php
+++ b/tests/unit/Interactions/Internal/WebDriverCoordinatesTest.php
@@ -2,6 +2,7 @@
 
 namespace Facebook\WebDriver\Interactions\Internal;
 
+use Facebook\WebDriver\WebDriverPoint;
 use PHPUnit\Framework\TestCase;
 
 class WebDriverCoordinatesTest extends TestCase
@@ -9,27 +10,16 @@ class WebDriverCoordinatesTest extends TestCase
     public function testConstruct()
     {
         $in_view_port = function () {
+            return new WebDriverPoint(0, 0);
         };
         $on_page = function () {
+            return new WebDriverPoint(10, 10);
         };
 
         $webDriverCoordinates = new WebDriverCoordinates(null, $in_view_port, $on_page, 'auxiliary');
 
-        $this->assertAttributeEquals(null, 'onScreen', $webDriverCoordinates);
-        $this->assertAttributeEquals($in_view_port, 'inViewPort', $webDriverCoordinates);
-        $this->assertAttributeEquals($on_page, 'onPage', $webDriverCoordinates);
-        $this->assertAttributeEquals('auxiliary', 'auxiliary', $webDriverCoordinates);
-    }
-
-    public function testGetAuxiliary()
-    {
-        $in_view_port = function () {
-        };
-        $on_page = function () {
-        };
-
-        $webDriverCoordinates = new WebDriverCoordinates(null, $in_view_port, $on_page, 'auxiliary');
-
-        $this->assertEquals('auxiliary', $webDriverCoordinates->getAuxiliary());
+        $this->assertEquals(new WebDriverPoint(0, 0), $webDriverCoordinates->inViewPort());
+        $this->assertEquals(new WebDriverPoint(10, 10), $webDriverCoordinates->onPage());
+        $this->assertSame('auxiliary', $webDriverCoordinates->getAuxiliary());
     }
 }
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 * text=auto
 
+/scripts export-ignore
 /tests export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
--- a/.github/workflows/php.yaml
+++ b/.github/workflows/php.yaml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Setup PHP
-        uses: shivammathur/setup-php@v1
+        uses: shivammathur/setup-php@v2
         with:
           php-version: '7.4'
           extensions: mbstring, intl, zip
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ composer.phar
 composer.lock
 vendor
 .php_cs.cache
+.phpunit.result.cache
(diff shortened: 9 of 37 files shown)


phpunit.xml.dist
```diff
@@ -18,11 +18,11 @@
         </testsuite>
     </testsuites>
 
-    <filter>
-        <whitelist>
+    <coverage processUncoveredFiles="true">
```
williamdes, line 21: This change is not necessary and is not good because we still support phpunit 5 ;)

driesvints: Yeah, I'm trying to get the PHP 8 build to pass first. I'll revert everything that's unnecessary of course :thu

driesvints:
Currently waiting on https://github.com/sminnee/phpunit-mock-objects/pull/6 to be merged and tagged before continuing here.

phpunit.xml.dist
sminnee: We had this issue on Silverstripe and I addressed this by creating sminnee/phpunit, a replacement that ships 5.7.30 with PHP8 support.

Ditto for sminnee/phpunit-mock-objects

You're welcome to use it if you want.

driesvints: Thanks @sminnee, I'll take a look at everything on Monday 👍

sminnee:
You may wish to run your php 8 build with --ignore-platform-reqs. Without dropping PHP 5.6 and 7.0 support, any upgrading to at least phpunit 7 || 8 || 9, you'll likely run into issues.

If you're looking to patch my phpunit forks, please test the new packages against phpunit/phpunit first, and confirm that the tests pass - this is what I got blocked on trying to upgrade some of the dependencies.

phpunit.xml.dist
driesvints: Reverted this. And implemented sminnee/phpunit, thanks!

driesvints:
@sminnee I've added "sminnee/phpunit" which indeed works thanks.

I'm a bit stuck atm... For some bizar reason when I only run the build for PHP 8 it succeeds (see [this build](https://github.com/php-webdriver/php-webdriver/runs/1342588938) and my second to last commit) but when I re-enable all the other builds it fails and claims it cannot install PHP CS Fixer even though I've added `--ignore-platform-req php` (see [this build](https://travis-ci.com/github/php-webdriver/php-webdriver/jobs/424113941) and my last commit). Apparently for some reason it seems that parallel builds in Travis influence each other somehow? In any case the PHP 8 nightly build should properly install because of the `--ignore-platform-req php` flag.

driesvints:
Also the PHP Code Style Github Action fails atm but I suspect that has nothing to do with the current PR (since it's run on PHP 7.4). I believe when you trigger it on main it'll also fail. Probably something got changed on phpstan's side.

driesvints:
I think I'm a little closer at figuring it out. As you can see below there's two nightly builds, one that fails and one that passes. I'm sure that the one that passes has the `--ignore-platform-req php` and the one that fails doesn't.

There's two questions:

- Why are there two builds?
- How do we get the `--ignore-platform-req php` to apply on the one that fails?

Screenshot 2020-11-02 at 17 37 30

.travis.yml
```diff
@@ -8,6 +8,7 @@ php:
   - '7.2'
   - '7.3'
   - '7.4'
+  - 'nightly'
```
williamdes, line 11: Remove this one and you will not have 2 builds ;)

driesvints: That seem to have done it. Thank you @williamdes! I guess I was confused here because if you leave it in place but disable all the other matrix builds it only runs one. That's pretty confusing...

williamdes: In fact if you add this one here it will run the matrix job for each value
And you added a special block for a job and that made it to be duplicated.
Here is an example of such override https://github.com/phpmyadmin/sql-parser/blob/37c4baeddee8f749b0ce3a26e88988c5d39dfa60/.travis.yml#L42

driesvints:
Okay this one is ready for review now. Please note that I had to disable code coverage for PHP 8 for now because of an issue with PHPUnit and xdebug. See https://travis-ci.com/github/php-webdriver/php-webdriver/builds/197177332

This was fixed in a newer version of PHPUnit later on but I guess `sminnee/phpunit` doesn't has that one yet (cc @sminnee). I'm however a bit too tired to figure out that specific one and don't believe it should be a blocker for this PR.

In general I want to suggest for this project to drop support for older PHP versions so you can remove the PHPUnit dependency overrides. As most of these PHP versions are EoL I don't think it makes sense in maintaining them anymore. However, I understand that this is a personal choice of the maintainer of course.

Anyway, would appreciate a review, merge and tag so we can finally patch Laravel Dusk. Let me know if anything else needs to be done. Thanks all! 🙂

.travis.yml
```diff
@@ -117,7 +122,8 @@ script:
   - if [ "$BROWSER_NAME" = "firefox" ]; then EXCLUDE_GROUP+="exclude-firefox,"; fi
   - if [ "$BROWSER_NAME" = "chrome" ]; then EXCLUDE_GROUP+="exclude-chrome,"; fi
   - if [ -n "$EXCLUDE_GROUP" ]; then EXTRA_PARAMS+=" --exclude-group $EXCLUDE_GROUP"; fi
-  - ./vendor/bin/phpunit --coverage-clover ./logs/coverage-clover.xml $EXTRA_PARAMS
+  - if [ "$TRAVIS_PHP_VERSION" == "nightly" ]; then ./vendor/bin/phpunit $EXTRA_PARAMS; fi
+  - if [ "$TRAVIS_PHP_VERSION" != "nightly" ]; then ./vendor/bin/phpunit --coverage-clover ./logs/coverage-clover.xml $EXTRA_PARAMS; fi
```
williamdes, line 126: is this temporary ?

composer.json
```diff
@@ -26,10 +26,10 @@
     "friendsofphp/php-cs-fixer": "^2.0",
     "ondram/ci-detector": "^2.1 || ^3.5",
     "php-coveralls/php-coveralls": "^2.0",
-    "php-mock/php-mock-phpunit": "^1.1",
+    "php-mock/php-mock-phpunit": "^1.1 || ^2.0",
     "php-parallel-lint/php-parallel-lint": "^1.2",
-    "phpunit/phpunit": "^5.7",
-    "sebastian/environment": "^1.3.4 || ^2.0 || ^3.0",
+    "sminnee/phpunit": "^5.7",
```
williamdes, line 31: we should stick to `phpunit/phpunit` and add new supported versions if possible.

.travis.yml
williamdes: --no-coverage could be enough to work

composer.json
driesvints: That's not possible because `sminnee/phpunit-mock-objects` is blocking that atm. See https://github.com/sminnee/phpunit-mock-objects/pull/6 and his reply (and his other replies on this PR).

driesvints:
For some bizar reason the PHP 5.6 til 7.2 builds are now failing after my last commit... https://github.com/php-webdriver/php-webdriver/pull/828/checks?check_run_id=1343276205

composer.json
williamdes: Okay, I can finish your work then ;)
Please give me access to your repository

driesvints: Not sure which one you meant so I gave you access to this one and the phpunit-mock-objects fork :)

.travis.yml
driesvints: This worked 👍

.travis.yml
driesvints: Right. But I think the current state of the PR is good enough?

composer.json
williamdes: Thank you, perfect :rocket: 
This was what I expected :)

williamdes:
Part of the build passes, I will copy my hack friends into the CI to make tests pass on modern versions of phpunit ;)
Source: https://salsa.debian.org/phpmyadmin-team/phpmyadmin/-/blob/debian/4%254.9.7+dfsg1-1_bpo10+1/debian/tests/phpunit#L17

.php_cs.dist
```diff
@@ -52,7 +52,7 @@ return PhpCsFixer\Config::create()
         'ordered_class_elements' => true,
         'ordered_imports' => true,
         'php_unit_construct' => true,
-        'php_unit_dedicate_assert' => true,
+        'php_unit_dedicate_assert' => false,
```
williamdes, line 55: Because assertInternalType is deprecated

williamdes:
Hi @OndraM !
This PR started with a PHP 8 support and I finished it with a phpunit 5.7->9 support.
Because only the CI runs phpunit tests on PHP 5 I did set a script block to reverse some method signatures like I do on Debian (https://github.com/php-webdriver/php-webdriver/pull/828#issuecomment-720622743).

I upgraded things and did fix everything that was red.
Some fixes/hacks are needed and can not be removed. But we can merge this and remove those things when we drop php 5 support.

driesvints:
@williamdes fantastic work 🙌

.php_cs.dist
GrahamCampbell: Just delete this line.

GrahamCampbell (changes requested):

williamdes:
Hi @GrahamCampbell 
> Just delete this line.

I can not do that because phpcs will report me to use some phpunit asserts that are not right.
See: 26c6de1cdb8575ac4d06f8c38530bb11c97d644e
Deprecated: https://github.com/sebastianbergmann/phpunit/issues/3369

mauriciofauth (approved):

tests/unit/Interactions/Internal/WebDriverCoordinatesTest.php
```diff
@@ -15,10 +16,19 @@ public function testConstruct()
 
         $webDriverCoordinates = new WebDriverCoordinates(null, $in_view_port, $on_page, 'auxiliary');
```
OndraM, line 17: This test does seem to focus on internal implementation... I know its unrelated to PHP 8, but instead of using reflectioins, we can rather replace it with something more useful, so I suggest replacing  whole method it with following:

```php
public function testConstruct()
    {
        $in_view_port = function () {
            return new WebDriverPoint(0, 0);
        };
        $on_page = function () {
            return new WebDriverPoint(10, 10);
        };

        $webDriverCoordinates = new WebDriverCoordinates(null, $in_view_port, $on_page, 'auxiliary');

        $this->assertEquals(new WebDriverPoint(0, 0), $webDriverCoordinates->inViewPort());
        $this->assertEquals(new WebDriverPoint(10, 10), $webDriverCoordinates->onPage());
        $this->assertSame('auxiliary', $webDriverCoordinates->getAuxiliary());
    }
```

Then we can also drop following "testGetAuxiliary" method.

.travis.yml
```diff
@@ -80,6 +93,15 @@ install:
   - travis_retry composer update --no-interaction $DEPENDENCIES
 
 before_script:
+  - if [ "$RUN_PHPUNIT_PATCHES" = "true" ]; then
```
OndraM, line 96: What about moving this to separate shell script to make the config shorter?

Also we may eg. soon migrate to GH Actions only, so there won't be a need to copy-paste it.

tests/unit/Interactions/Internal/WebDriverCoordinatesTest.php
driesvints: Done 👍

tests/functional/WebDriverTestCase.php
```diff
@@ -213,4 +213,20 @@ protected function setUpSauceLabs()
...
         }
     }
+
+    /**
+     * Uses assertStringContainsString when it is available or uses assertContains for old phpunit versions
+     * @param string $needle
+     * @param string $haystack
+     * @param string $message
+     */
+    protected function compatAssertStringContainsString($needle, $haystack, $message = '')
```
OndraM, line 223: Just in idea, isn't possible to name it simply "assertStringContainsString", and call the parent if it exists, or assertContains if not?

Once we drop old PHPUnit versions, we could then only remove this polyfill.

driesvints: @williamdes can you maybe check in on this one?

.travis.yml
driesvints: @williamdes can you maybe check in on this one?

tests/functional/WebDriverTestCase.php
williamdes: No I can not do that, neither.
Because if I rename to `compatAssertStringContainsString` I will have to be compatible on the PHP signature, that would be one particular phpunit version. And so I would screw all the other ones.
Some not supporting phptype hints, some having some typehints defined, some no typehints defined.
:man_shrugging:

OndraM: Though the signature removal is already being done in via `sed` for similar cases like `setUpBeforeClass` etc., so theoretically it could be possible for `assertStringContainsString` signature as well, no? 

However it was just an idea, if you think its not worth it, we can keep it like this :).

williamdes: Well it is possible but totally overkill, let's keep it as it is ;)
I say it is overkill because we would need to make different seds for different phpunit versions and this is more complicated than just switching `: void` return typehints.
So is this okay for you until we drop some phpunit versions ?

.travis.yml
williamdes: Done in 720c2750d4eb3064e6363f3f7a7c16fba75cca59

driesvints: Just a really minor thing but maybe name the directory `bin/` so it's more in line with other composer libraries?

williamdes: I would say no, on [Doctum](https://github.com/code-lts/doctum/tree/main/bin) and some other libs at [phpMyAdmin](https://github.com/phpmyadmin/sql-parser/tree/master/bin) scripts found in bin are for users of the lib :)

driesvints: Gotcha 👍

OndraM:
LGTM, thanks everyone!

The final question is how to rebase & fixup those 25 commits to keep the changes contained. Any preferences @driesvints  @williamdes?
I suggest keeping just two commits, one which adds Travis builds, second one with all the changes from @williamdes :).

driesvints:
@OndraM all good to me. I'll leave it up to @williamdes to decide on that since he did the bulk of the work here 👍

williamdes:
> @OndraM all good to me. I'll leave it up to @williamdes to decide on that since he did the bulk of the work here

Done, I did a reset soft and created 2 commits from the changes.
And added `Dries Vints <dries@vints.io>` as a co author on 06ff635be7be7fc3ca20f8f4bf2406e7324236a3

driesvints:
@williamdes can you try to retrigger the static analysis github action? It might just be a hiccup.

williamdes:
Hi @ondrejmirtes
This phpstan error seems to be recent, is it an internal bug ?
https://github.com/php-webdriver/php-webdriver/runs/1406254306?check_suite_focus=true#step:6:28

ondrejmirtes:
@williamdes https://github.com/phpstan/phpstan/issues/3956

OndraM:
So maybe we can temporarily add constrain to composer.json to avoid latest PHPStan version(s)?

ondrejmirtes:
It'd be easier to:

1) Avoid using non-UTF-8 PHP code in your project. It's 2020 :)
2) The root issue is reported here (https://github.com/clue/reactphp-ndjson/issues/21). I don't know what to do about it right now. I don't know anything about the encoding of the analysed codebse so I don't know what to convert it from to UTF-8.

OndraM:
@ondrejmirtes AFAIK this whole codebase is UTF-8. Is there a way how to find out which file causes the trouble?

I also tried this and find no non-UTF-8 files... https://gist.github.com/rmetzler/2947828

williamdes:
I tried on my local workstation:
```
composer analyse
# fails
vendor/bin/phpstan.phar analyze ./lib ./tests --level 2 -c phpstan.neon --ansi
# fails
vendor/bin/phpstan.phar analyze ./lib ./tests --level 2 -c phpstan.neon --ansi
# fails
vendor/bin/phpstan.phar analyze ./lib ./tests --level 2 -c phpstan.neon --ansi --debug -vvv
# SUCCESS
vendor/bin/phpstan.phar analyze ./lib ./tests --level 2 -c phpstan.neon --ansi
# SUCCESS, WTF
rm -rf /tmp/phpstan
vendor/bin/phpstan.phar analyze ./lib ./tests --level 2 -c phpstan.neon --ansi
# fails
```

@ondrejmirtes `--debug -vvv` was added in the latest release, and before this release the analysis was successing.
Is it by coincidence this new error ?

ondrejmirtes:
The reason why it fails are these two commits:

* https://github.com/phpstan/phpstan-src/commit/95101a01350698ecaa447fbf791516b8f0a5ccee
* https://github.com/phpstan/phpstan-src/commit/14928bf41fbe80ba68ab484e138138442d1cef4a

So it's not a new bug, it just surfaces an error condition that was failing silently before, possibly causing error reports to be missed.

driesvints:
@OndraM @williamdes is there any way we can work around this for now so we can merge this PR? Afaik this has nothing to do with this PR and it'll probably also fail on the main branch?

OndraM:
From what I found, the cause is in [WebDriveKeys](https://github.com/php-webdriver/php-webdriver/blob/main/lib/WebDriverKeys.php#L12-L81). However we need to declare WebDriver key codes there... Doing this is nothing invalid from PHP point of view. @ondrejmirtes So this is caused by https://github.com/clue/reactphp-ndjson/issues/21 , which is used by PHPStan, and which causes non-unicode characters to break JSON parsing used internally in PHPStan?

OndraM:
> @OndraM @williamdes is there any way we can work around this for now so we can merge this PR? Afaik this has nothing to do with this PR and it'll probably also fail on the main branch?

Yes, will probably do this. I just wanted to know what is the root cause for this and whether it is related to this PR.

OndraM:
Workaround for the PHPStan issue here: https://github.com/php-webdriver/php-webdriver/pull/830

phpstan.neon
```diff
@@ -7,8 +7,6 @@ parameters:
     ignoreErrors:
         # To be fixed:
         - '#Call to an undefined method Facebook\\WebDriver\\WebDriver::getTouch\(\)#'
-        - '#Call to an undefined method Facebook\\WebDriver\\WebDriverElement::getCoordinates\(\)#'
-        - '#Call to an undefined method Facebook\\WebDriver\\WebDriverElement::equals\(\)#'
```
OndraM, line 11: And now those needs to be put back :upside_down_face: .

williamdes: Thank you !
Happy you did point that out because I was searching why the error was occurring ^^

driesvints:
@ondrejmirtes @OndraM @williamdes thanks all! 🙂

OndraM:
PHP 8 support is now merged to `main` branch. Nice job everyone! 

I will get this released as `1.8.4` quite soon.

viezel:
@ondrejmirtes @OndraM @williamdes awesome work. Looking forward to that `1.9.0` tag! 
PHP8 here we go! 🚀

driesvints:
Hey @OndraM. Is there anything blocking the 1.9.0? We'd really like to patch Dusk soon since it's currently blocking a lot of other projects from shipping PHP 8 support. Let me know if there's anything I can do to help out.

OndraM:
@driesvints Sorry for the delay, I'd like to finish one other PR (#556) to get it merged into the 1.9.0 release. Should be finished once CI builds are done, and the release will be just afterwards :)

driesvints:
@OndraM no sorry needed. Awesome, looking forward to it :)

OndraM:
:loudspeaker: 1.9.0 with PHP 8 support was just released.

driesvints:
@OndraM amazing!

scripts/apply-phpunit-patches.sh
```diff
@@ -0,0 +1,20 @@
...
+find tests/ -type f -print0 | xargs -0 sed -i 's/function tearDown(): void/function tearDown()/g';
+
+sed -i 's/endTest(\\PHPUnit\\Framework\\Test \$test, float \$time): void/endTest(\\PHPUnit_Framework_Test \$test, \$time)/g' tests/functional/ReportSauceLabsStatusListener.php;
+sed -i 's/: void/ /g' tests/functional/ReportSauceLabsStatusListener.php;
+# Drop the listener from the config file
+sed -i '/<listeners>/,+2d' phpunit.xml.dist;
+sed -i 's/function runBare(): void/function runBare()/g' tests/functional/WebDriverTestCase.php;
+
+# Return back to original dir
+cd - > /dev/null
```
driesvints, line 20: @OndraM @williamdes would it be okay with you two if I used this script for https://github.com/algolia/algoliasearch-client-php/pull/644 as well?

williamdes: Yes you can copy it :)

OndraM: 👍

PR #832, opened 2020-11-20
Execute Chrome DevTool Protocol (CDP) commands

OndraM:
This adds ability to execute Chrome DevTool Protocol (CDP) commands via Chromedriver HTTP endpoint.

Unlike in previous implementation in #556 , access to CDP is encapsulated to stand-alone class `ChromeDevToolsDriver`, which provide access to CDP not only from ChromeDriver instance (= Chrome running on the local machine), but also when Chrome is running remotely via Selenium.

```php
// local Chrome
$driver = ChromeDriver::start();

$devTools = $driver->getDevTools(); // return ChromeDevToolsDriver instance
$devTools->execute('Performance.enable');
```
```php
// remote Chrome
$driver = RemoteWebDriver::create('...', DesiredCapabilities::chrome());

$devToolsDriver = new ChromeDevToolsDriver($driver);
$devToolsDriver->execute('Performance.enable');
```

Class `ChromeDevToolsDriver` is open for possible future extension (there could be for example some abstraction or syntax sugar for the most common CDP commands etc.).

"Official" Java bindings do however use `driver.executeCdpCommand()` method (similar to what was proposed in #556), but this comes with the limitation mentioned above, so I think this solution is more flexible. See documentation for [Chrome DevTools in Java bindings](https://www.selenium.dev/documentation/en/support_packages/chrome_devtools/).

### Note for possible use-cases
Because CDP commands are executed on top of WebDriver HTTP protocol, it does not provide the benefits of CDP control directly via WebSockets (this is possible using [dedicated libraries](https://github.com/jakubkulhan/chrome-devtools-protocol) - but works only with local Chrome and have other limitations). 

However it may still be quite useful in case you need only some simple Chrome quirks (like changing user agent, manipulating geolocation etc.), and you still want all features and benefits of Selenium and WebDriver. 

(BTW there is also proposed [WebDriver BiDi protocol](https://w3c.github.io/webdriver-bidi/), with attempt of standardizing some parts of CDP to make them cross-browser compatible.)

### After merge ###
- [ ] Update [documentation](https://github.com/php-webdriver/php-webdriver/wiki/)
- [ ] Update changelog

### DIFF
--- a/lib/Chrome/ChromeDevToolsDriver.php
+++ b/lib/Chrome/ChromeDevToolsDriver.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Facebook\WebDriver\Chrome;
+
+use Facebook\WebDriver\Remote\RemoteWebDriver;
+
+/**
+ * Provide access to Chrome DevTools Protocol (CDP) commands via HTTP endpoint of Chromedriver.
+ *
+ * @see https://chromedevtools.github.io/devtools-protocol/
+ */
+class ChromeDevToolsDriver
+{
+    const SEND_COMMAND = [
+        'method' => 'POST',
+        'url' => '/session/:sessionId/goog/cdp/execute',
+    ];
+
+    /**
+     * @var RemoteWebDriver
+     */
+    private $driver;
+
+    public function __construct(RemoteWebDriver $driver)
+    {
+        $this->driver = $driver;
+    }
+
+    /**
+     * Executes a Chrome DevTools command
+     *
+     * @param string $command The DevTools command to execute
+     * @param array $parameters Optional parameters to the command
+     * @return array The result of the command
+     */
+    public function execute($command, array $parameters = [])
+    {
+        $params = ['cmd' => $command, 'params' => (object) $parameters];
+
+        return $this->driver->executeCustomCommand(
+            self::SEND_COMMAND['url'],
+            self::SEND_COMMAND['method'],
+            $params
+        );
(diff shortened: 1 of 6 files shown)


lib/Chrome/ChromeDevToolsDriver.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Executes a Chrome DevTools command
+     *
+     * @param string $command The DevTools command to execute
+     * @param array $parameters Optional parameters to the command
+     * @return array The result of the command
+     */
+    public function execute($command, array $parameters = [])
```
williamdes, line 36:
```suggestion
    public function execute(string $command, array $parameters = [])
```

williamdes (approved):

lib/Chrome/ChromeDevToolsDriver.php
OndraM: No :disappointed:

williamdes: :man_shrugging: One day :smile:

OndraM:
I've completely revamped our wiki with Chrome information.

It now includes various CDP examples:
- Change user-agent
- Get performance metrics
- Emulate geolocation
- Set custom request header


See https://github.com/php-webdriver/php-webdriver/wiki/Chrome#chrome-devtools-protocol-cdp

PR #841, opened 2020-12-06
Put more information to every deprecation, deprecate PhantomJS

OndraM:
Its quite helpful for library users to have comment with upgrade path or with a link to more information on every deprecation.

Also, we can mark PhantomJS deprecated (it is no longer developed, headless Chrome or Firefox should be used instead), and its support from the code will be completely removed in next major version.


williamdes (commented):
LGTM

Issue #849, opened 2020-12-27
ChromeOptions::CAPABILITY_W3C is ignored in ChromeDriver


gvlasov:
```php
$chromeOptions = new ChromeOptions();
$chromeOptions->addArguments([
    '--headless',
    // or whatever arguments
]);
$capabilities->setCapability(
     ChromeOptions::CAPABILITY_W3C,
     $chromeOptions
);
```

Doesn't pass "goog:chromeOptions" capability to Chromedriver node when establishing a connection, because this code:

https://github.com/php-webdriver/php-webdriver/blob/main/lib/Remote/DesiredCapabilities.php#L169-L186

Doesn't try to cast `ChromeOptions` object under index `ChromeOptions::CAPABILITY_W3C` to an array.

The workaround is:

```php
$chromeOptions = new ChromeOptions();
$chromeOptions->addArguments([
     '--headless',
     // or whatever arguments
]);
$capabilities->setCapability(
     ChromeOptions::CAPABILITY_W3C,
     $chromeOptions->toArray() // <<<<<<<<<<<<<<<<<<<<<<<<<<
);
```
but that's like super unintuitive, ok.

OndraM:
@gvlasov Hi, thanks for the report, however, removing the prepared template for bug report make it for us much harder to take action.

What are the steps you use to start the browser? Could you please provide entire code you use to initialize the browser (not just the snippet with capabilities).

gvlasov:
@OndraM You know these details you're asking for don't matter. This is just a missed case in library serialization code.

OndraM:
These information matter, because you didn't provide complete self-contained and reproducible bug report - which is the reason why we require the bug template is properly and completely filled. Please understand proper bug reports help us quickly understand when the bug appears and how we can fix it. Without these information, we can only guess - which could lead to not fixing the issue properly.

gvlasov:
@OndraM Dude, everyone but you understands this bug, it is trivial if you actually read the report and not whine about it not conforming to some standard. I provided enough info and there are already changes being made from your team members who understand the issue. You seem to be attached to your bureaucracy a little too much and overestimate my level of investment in your library.

OndraM:
This is not about bureaucracy, but about effectiveness and increasing a chance of finding a solution to the issue.

[...]

PR #850, opened 2020-12-27
Serialize ChromeOptions object under ChromeOptions::CAPABILITY_W3C capability index

gvlasov:
For bug report #849

### DIFF
--- a/lib/Remote/DesiredCapabilities.php
+++ b/lib/Remote/DesiredCapabilities.php
@@ -175,6 +175,13 @@ class DesiredCapabilities implements WebDriverCapabilities
                 $this->capabilities[ChromeOptions::CAPABILITY]->toArray();
         }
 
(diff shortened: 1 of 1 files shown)


gvlasov:
These changes should have been present in https://github.com/php-webdriver/php-webdriver/pull/673

williamdes (approved):

andrewnicols:
I don't think that this is the right approach. The current approach there is a stop-gap solution which has side-effects as noted in the TODO.

See my alternative proposal in #852

andrewnicols:
I think that we can close this issue now that #852 has landed. Landing this change is not mutually exclusive, but will lead to further b/c breakages in future when we do release 2.0 and stop converting modifying the content of `$this->capabilities`.

Do you agree @OndraM?

OndraM:
Sure, closing this, as it has been covered by #852. Thanks @gvlasov for the suggestion, though!

PR #852, opened 2020-12-30
Make the capability conversion JsonSerializable

### LINKED ISSUE #849: ChromeOptions::CAPABILITY_W3C is ignored in ChromeDriver

andrewnicols:
By making the ChromeOptions and FirefoxProfile implement
JsonSerializable, both the ChromeOptions::CAPABILITY and
ChromeOptions::CAPABILITY_W3C will be serializable when converting to
JSON for the Desired Capabilities.

This also solves an issue whereby it is not possible to specify
CAPABILITY_W3C as the capability key.

Fixes #849

### DIFF
--- a/lib/Chrome/ChromeOptions.php
+++ b/lib/Chrome/ChromeOptions.php
@@ -3,13 +3,14 @@
 namespace Facebook\WebDriver\Chrome;
 
 use Facebook\WebDriver\Remote\DesiredCapabilities;
+use JsonSerializable;
 
 /**
  * The class manages the capabilities in ChromeDriver.
  *
  * @see https://sites.google.com/a/chromium.org/chromedriver/capabilities
  */
-class ChromeOptions
+class ChromeOptions implements JsonSerializable
 {
     /**
      * The key of chrome options desired capabilities (in legacy OSS JsonWire protocol)
@@ -36,6 +37,16 @@ class ChromeOptions
      * @var array
      */
     private $experimentalOptions = [];
+
+    /**
+     * Return a version of the class which can JSON serialized.
+     *
+     * @return array
+     */
+    public function jsonSerialize()
+    {
+        return $this->toArray();
+    }
 
     /**
      * Sets the path of the Chrome executable. The path should be either absolute
--- a/tests/unit/Remote/DesiredCapabilitiesTest.php
+++ b/tests/unit/Remote/DesiredCapabilitiesTest.php
@@ -143,9 +143,9 @@ class DesiredCapabilitiesTest extends TestCase
         DesiredCapabilities $inputJsonWireCapabilities,
         array $expectedW3cCapabilities
     ) {
-        $this->assertEquals(
-            $expectedW3cCapabilities,
-            $inputJsonWireCapabilities->toW3cCompatibleArray()
+        $this->assertJsonStringEqualsJsonString(
+            json_encode($expectedW3cCapabilities),
+            json_encode($inputJsonWireCapabilities->toW3cCompatibleArray())
         );
     }
 
@@ -224,6 +224,18 @@ class DesiredCapabilitiesTest extends TestCase
             'chromeOptions should be converted' => [
                 new DesiredCapabilities([
                     ChromeOptions::CAPABILITY => $chromeOptions,
+                ]),
+                [
+                    'goog:chromeOptions' => new \ArrayObject(
+                        [
+                            'args' => ['--headless'],
+                        ]
+                    ),
+                ],
+            ],
+            'chromeOptions as W3C capability should be converted' => [
+                new DesiredCapabilities([
+                    ChromeOptions::CAPABILITY_W3C => $chromeOptions,
                 ]),
                 [
                     'goog:chromeOptions' => new \ArrayObject(


andrewnicols:
Since there's no interface to handle array casting, we have two options:
1. handle the cast in the json_encode() call in HttpExecutor
2. call something like `return (array) json_decode(json_encode($this->capabilities));` in `DesiredCapabilities`.

Any thoughts @OndraM?

williamdes (approved):

lib/Remote/DesiredCapabilities.php
```diff
@@ -163,25 +163,10 @@ public function setJavascriptEnabled($enabled)
     }
 
     /**
-     * @todo Remove side-effects - not change eg. ChromeOptions::CAPABILITY from instance of ChromeOptions to an array
      * @return array
      */
     public function toArray()
     {
-        if (isset($this->capabilities[ChromeOptions::CAPABILITY]) &&
```
oleg-andreyev, line 171: the problem with these changes that it's a B/C break, it's changing the shape of a return, before this PR "ChromeOptions::CAPABILITY" and "FirefoxDriver::PROFILE" are converted into an array but after this PR they will be returned as is.

OndraM (commented):
I agree with @oleg-andreyev that removing the code from `toArray()` may be a BC break for someone, because unfortunately someone can rely on the side-effects. I already looked into this in https://github.com/php-webdriver/php-webdriver/pull/643#issuecomment-500789115 :confused:  

But if I understand the issue from #849 correctly, this could be fixed with just the JsonSerializable interface added to ChromeOptions and Firefox but with no other change - because the value of capabilities is in the end converted to json in [HttpCommandExecutor](https://github.com/php-webdriver/php-webdriver/blob/main/lib/Remote/HttpCommandExecutor.php#L311).

(However, for 2.0 I think `toArray()` should be changed to be immutable without side-effects).

What do you think, @andrewnicols ?

andrewnicols:
Okay, so how about this:

1. Add a new `Facebook\WebDriver\BrowserOptions` interface which defines a function `public function toCapabilities()` which returns mixed
2. Implement this in ChromeOptions as `return $this->toArray();` (returns an array)
3. Implement it in FirefoxProfile as `return $this->encode()` (returns a string)
4. Modify the DesiredCapabilites::toArray() function to something like:
```
public function toArray()
{
    $capabilities = [];
    foreach ($this->capabilities as $key => $capability) {
        if ($capability instanceof BrowserOptions) {
            $capabilities[$key] = $capability->toCapabilities();
        } else {
        
        }
    }
}
```
5.

andrewnicols:
Okay, I think I understand the suggestion an I've made the requested changes.

lib/Remote/DesiredCapabilities.php
andrewnicols: Arguably this is correct and the fact that they return objects is a bug, but I agree that this would be a change in behaviour and necessity a major version bump.

lib/Remote/DesiredCapabilities.php
```diff
@@ -237,8 +237,9 @@ public function toW3cCompatibleArray()
             if (array_key_exists(ChromeOptions::CAPABILITY_W3C, $ossCapabilities)) {
                 $w3cCapabilities[ChromeOptions::CAPABILITY_W3C] = new \ArrayObject(
                     array_merge_recursive(
-                        (array) $ossCapabilities[ChromeOptions::CAPABILITY],
-                        (array) $ossCapabilities[ChromeOptions::CAPABILITY_W3C]
+                        // Encode and decode from JSON to cast the value to an array.
+                        (array) json_decode(json_encode($ossCapabilities[ChromeOptions::CAPABILITY])),
+                        (array) json_decode(json_encode($ossCapabilities[ChromeOptions::CAPABILITY_W3C]))
```
OndraM, line 242: Is this still necessary? The object is jsonserializable, so should be converted to array automatically by json_encode right before passing the value to curl.

andrewnicols: I’ll double-check but I think some of the unit tests were failing without this. I can roll this back and fix it in the unit tests instead if that’s the case. Will check shortly.

andrewnicols: Tests are passing without. This may have been required before because of the way that I was performing the changes prior to the updaet.

lib/Chrome/ChromeOptions.php
```diff
@@ -37,6 +38,16 @@ class ChromeOptions
      */
     private $experimentalOptions = [];
 
+    /**
+     * Return a version of the class which can JSON serialized.
+     *
+     * @return  array
```
OndraM, line 44:
```suggestion
     * @return array
```

lib/Firefox/FirefoxProfile.php
```diff
@@ -27,6 +28,16 @@ class FirefoxProfile
      */
     private $rdf_file;
 
+    /**
+     * Return a version of the class which can JSON serialized.
+     *
+     * @return  array
```
OndraM, line 34:
```suggestion
     * @return array
```

lib/Firefox/FirefoxProfile.php
```diff
@@ -27,6 +28,16 @@ class FirefoxProfile
...
+     */
+    public function jsonSerialize()
+    {
+        return $this->encode();
```
OndraM, line 38: But this returns string - while it must return something serializable using `json_encode()`. It also means we don't have a test for this (otherwise it would fail). And we should increase phpstan level, the current level 2 one does not check for this 😬.

But do we actually need JsonSerializable on FirefoxProfile? I believe the issue in #849 is just with ChromeOptions and FirefoxProfile is not related.

tests/unit/Remote/DesiredCapabilitiesTest.php
```diff
@@ -143,9 +143,9 @@ public function testShouldConvertCapabilitiesToW3cCompatible(
         DesiredCapabilities $inputJsonWireCapabilities,
         array $expectedW3cCapabilities
     ) {
-        $this->assertEquals(
-            $expectedW3cCapabilities,
-            $inputJsonWireCapabilities->toW3cCompatibleArray()
+        $this->assertJsonStringEqualsJsonString(
```
OndraM, line 146: I suggest adding testcase below, which will use of JsonSerializable (and which would cover the scenario from #849).

Something like:
```
            'chromeOptions as W3C capability should be converted' => [
                new DesiredCapabilities([
                    ChromeOptions::CAPABILITY_W3C => $chromeOptions,
                ]),
                [
                    'goog:chromeOptions' => new \ArrayObject(
                        [
                            'args' => ['--headless'],
                        ]
                    ),
                ],
            ],
```

OndraM (changes requested):
Sorry for not catching these earlier :disappointed:, but could you please have look once more? :pray:

lib/Firefox/FirefoxProfile.php
andrewnicols: Yes - you're right. I misunderstood the purpose of FirefoxProfile. Killing that change.

tests/unit/Remote/DesiredCapabilitiesTest.php
andrewnicols: Done - this testcase sufficed with no need to make any other changes.

lib/Firefox/FirefoxProfile.php
```diff
@@ -4,6 +4,7 @@
 
 use Facebook\WebDriver\Exception\WebDriverException;
 use FilesystemIterator;
+use JsonSerializable;
```
oleg-andreyev, line 7: Looks like class is not used.

andrewnicols: Bah - good catch. Fixed.

OndraM (approved):
Nice, thanks!

PR #853, opened 2021-01-07
Pass W3C capabilitites when starting chromedriver

OndraM:
While looking into #849 and #852, I discovered we are not passing capabilities in W3C format when staring browser via `ChromeDriver::start()`, as we do [in RemoteWebDriver::create()](https://github.com/php-webdriver/php-webdriver/blob/main/lib/Remote/RemoteWebDriver.php#L108-L113). With this PR, capabilities as submitted in both formats.

Because chromedriver accepts capabilities in new session command in both formats, **there will be no effective change** in behavior right now. But should Chromedriver team decide to remove JsonWire protocol support, creating new browser session will not work without this change. By this change, we are adding **forward compatibility**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project versioning adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Support for sending Chrome DevTools Protocol commands (see details in [wiki](https://github.com/php-webdriver/php-webdriver/wiki/Chrome#chrome-devtools-protocol-cdp)).
 
+### Fixed
+- Actually start ChromeDriver in W3C mode if it is supported by the browser driver. Until now, when it was initialized using `ChromeDriver::start()`, it has always been unintentionally started in OSS mode.
+
 ### Changed
 - Throw `DriverServerDiedException` on local driver process terminating unexpectedly and provide full details of original exception to improve debugging.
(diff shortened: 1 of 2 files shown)


williamdes (approved):
LGTM

andrewnicols:
This change looks good to me. It matches the RemoteWebDriver::create() as expected.

I’m not sure why coverage has fallen as this shouldn’t lead to a decrease in covered lines but I’ve seen the coveralls tests showing things like this a bit recently so I wonder if it’s a false neg.

andrewnicols (approved):

OndraM:
> I wonder if it’s a false neg.

Yes, I noticed this as well - seems as a false positive 🤷

OndraM:
Oh no, actually, its not false positive! 😬 

I investigated a bit more and found that because we sent only JsonWire capabilities format, ChromeDriver **was in fact always started in w3c mode** (when stared via `ChromeDriver::start()`; when started via RemoteWebDriver, it was correctly in W3C mode by default). This is quite a surprising finding, considering no-one noticed for that long :smiley: . Chromium team is obviously doing a nice job with maintaining compatibility.

Because of this change, chrome was now started in W3C mode, and returned capabilities were in W3C format as well, so the line `$this->executor->disableW3cCompliance();` was now not executed (and this coverage drop).

Because this now in fact **change behavior**, I will add this to the changelog as a fix.

williamdes (approved):

andrewnicols (approved):

PR #854, opened 2021-01-12
Work around a bug in Geckodriver (653)

andrewnicols:
Hi @OndraM ,

This is a bit of a weird one, and I have fixed it in the Mink driver, but I feel that others may also benefit from it too.
I can't think of a better way to address this reliably.

This issue works around a long-standing bug in Geckodriver whereby if
the first child of an anchor is a block-level element, that element
covers the click-point of the anchor. This causes an
ElementNotInteractable exception to be thrown as Marionette believes
that it cannot be scrolled into view.

The workaround to this issue is to attempt to click on an immediate
child node instead. Usually the first child will be sufficient, but if
that first child is a hidden element, then a sibling should instead be
used.

This issue has existed for several years and there is no timeframe for
fixing it upstream.

The upstream issue is https://github.com/mozilla/geckodriver/issues/653 and the gecko issue is https://bugzilla.mozilla.org/show_bug.cgi?id=1374283. I have considered fixing this upstream but it's a crazy complicated problem to solve, I don't currently have time to dedicate to it, and @whimboo is keen to fix this in the spec before fixing it in geckodriver/marionette.

### DIFF
--- a/lib/Remote/RemoteWebElement.php
+++ b/lib/Remote/RemoteWebElement.php
@@ -2,6 +2,7 @@
 
 namespace Facebook\WebDriver\Remote;
 
+use Facebook\WebDriver\Exception\ElementNotInteractableException;
 use Facebook\WebDriver\Exception\WebDriverException;
 use Facebook\WebDriver\Interactions\Internal\WebDriverCoordinates;
 use Facebook\WebDriver\Internal\WebDriverLocatable;
@@ -69,10 +70,17 @@ class RemoteWebElement implements WebDriverElement, WebDriverLocatable
      */
     public function click()
     {
-        $this->executor->execute(
-            DriverCommand::CLICK_ELEMENT,
-            [':id' => $this->id]
-        );
+        try {
+            $this->executor->execute(
+                DriverCommand::CLICK_ELEMENT,
+                [':id' => $this->id]
+            );
+        } catch (ElementNotInteractableException $e) {
+            // An issue with geckodriver (https://github.com/mozilla/geckodriver/issues/653) prevents clicking on a link
+            // if the first child is a block-level element.
+            // The workaround in this case is to click on a child element.
+            $this->clickChildElement($e);
+        }
 
         return $this;
     }
@@ -475,6 +483,40 @@ class RemoteWebElement implements WebDriverElement, WebDriverLocatable
             ':id' => $this->id,
             ':other' => $other->getID(),
         ]);
+    }
+
+    /**
+     * Attempt to click on a child level element.
+     *
+     * This provides a workaround for geckodriver bug 653 whereby a link whose first element is a block-level element
+     * throws an ElementNotInteractableException could not scroll into view exception.
+     *
+     * The workaround provided here attempts to click on a child node of the element.
+     * In case the first child is hidden, other elements are processed until we run out of elements.
+     *
+     * @param ElementNotInteractableException $originalException The exception to throw if unable to click on any child
+     * @see https://github.com/mozilla/geckodriver/issues/653
+     * @see https://bugzilla.mozilla.org/show_bug.cgi?id=1374283
+     */
+    protected function clickChildElement(ElementNotInteractableException $originalException)
+    {
+        $children = $this->findElements(WebDriverBy::xpath('./*'));
+        foreach ($children as $child) {
+            try {
+                // Note: This does not use $child->click() as this would cause recursion into all children.
+                // Where the element is hidden, all children will also be hidden.
+                $this->executor->execute(
+                    DriverCommand::CLICK_ELEMENT,
+                    [':id' => $child->id]
+                );
+
+                return;
+            } catch (ElementNotInteractableException $e) {
+                // Ignore the ElementNotInteractableException exception on this node. Try the next child instead.
+            }
+        }
+
+        throw $originalException;
     }
 
     /**
--- a/tests/functional/RemoteWebElementTest.php
+++ b/tests/functional/RemoteWebElementTest.php
@@ -2,6 +2,7 @@
 
 namespace Facebook\WebDriver;
 
+use Facebook\WebDriver\Exception\ElementNotInteractableException;
 use Facebook\WebDriver\Exception\NoSuchElementException;
 use Facebook\WebDriver\Remote\RemoteWebElement;
 use Facebook\WebDriver\Remote\WebDriverBrowserType;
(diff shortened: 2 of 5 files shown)


lib/Remote/RemoteWebElement.php
```diff
@@ -69,10 +70,17 @@ public function clear()
...
-            [':id' => $this->id]
-        );
+        try {
+            $this->executor->execute(
+                DriverCommand::CLICK_ELEMENT,
+                [':id' => $this->id]
+            );
+        } catch (ElementNotInteractableException $e) {
+            // A long-standing bug in geckodriver/marionette prevents clicking on a link if the first child is a
+            // block-level element.
```
OndraM, line 80: What about adding reference to the bug in bugzilla? So that we can find the bug in the future to see if it is still present.

lib/Remote/RemoteWebElement.php
```diff
@@ -477,6 +485,39 @@ public function equals(WebDriverElement $other)
         ]);
     }
 
+    /**
+     * Attempt to click on a child level element.
+     *
+     * This provides a workaround for geckodriver bug 653 whereby a link whose first element is a block-level element
```
OndraM, line 491: Also rather full URL (maybe both github + bugzilla here) then just bug ID.

lib/Remote/RemoteWebElement.php
```diff
@@ -477,6 +485,39 @@ public function equals(WebDriverElement $other)
...
+     * throws an ElementNotInteractableException could not scroll into view exception.
+     *
+     * The workaround provided here attempts to click on a child node of the element.
+     * In case the first child is hidden, other elements are processed until we run out of elements.
+     *
+     * @param ElementNotInteractableException $originalException The exception to throw if unable to click on any child
+     * @return RemoteWebElement the current instance.
```
OndraM, line 498: I think there is no reason to provide fluent interface. This could just be kept as void.

tests/functional/web/links.html
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Links page</title>
```
OndraM, line 4: I think this whole could be named in reference of the gecko bug. Ie. not as generic "links test page", but page for testing of this specific scenario. 

We can also maybe omit the link index.html, as this is just this single-purpose page...

tests/functional/RemoteWebElementTest.php
```diff
@@ -117,6 +118,65 @@ public function testShouldClick()
...
 
+    /**
+     * Test that workarounds to geckodriver#653 works as expected.
+     *
+     * @covers ::click
+     * @covers ::clickChildElement
+     * @dataProvider provideValidClick
+     * @param string $id
+     */
+    public function testShouldClickOnBlockLevelElement($id)
```
OndraM, line 129: I know its generally bad idea in tests, but considering the time it takes to initialize new browser, I think the scales are a bit different with independency for integration tests of this kind - so would it be an issue by butting all the four cases into single test method?

tests/functional/RemoteWebElementTest.php
```diff
@@ -117,6 +118,65 @@ public function testShouldClick()
         );
     }
 
+    /**
+     * Test that workarounds to geckodriver#653 works as expected.
+     *
+     * @covers ::click
+     * @covers ::clickChildElement
+     * @dataProvider provideValidClick
+     * @param string $id
```
OndraM, line 127: Also makes it sense to limit the tests just to firefox?

tests/functional/RemoteWebElementTest.php
```diff
@@ -117,6 +118,65 @@ public function testShouldClick()
...
+        $linkElement = $this->driver->findElement(WebDriverBy::id($id));
+
+        $this->expectException(ElementNotInteractableException::class);
+        $linkElement->click();
+    }
+
+    /**
+     * @return array
+     */
+    public function provideInvalidClick()
```
OndraM, line 172: Dtto, let's make it single test?

OndraM (changes requested):

lib/Remote/RemoteWebElement.php
andrewnicols: I've added a @see link for both

tests/functional/web/links.html
andrewnicols: Yup - I  agree.

tests/functional/RemoteWebElementTest.php
andrewnicols: I'll add `@group exclude-chrome` to the test.

tests/functional/RemoteWebElementTest.php
andrewnicols: This one is more of a pain because of the try/catch but there's only two test conditions so easy enough.

andrewnicols:
We didn't have an exclude group for htmlunit so I've added that to .travis.yml
I've also added a new `skipForUnmatchedBrowsers()` fn to the functional testcase since the exclude groups are not added automatically if running locally and this test fails on HtmlUnit because the exception does not match (it's a `Facebook\WebDriver\Exception\UnrecognizedExceptionException` which probably needs its own bug anyway).

andrewnicols:
Okay, I think I have all of the exclusions right now:
* in phpdoc for the affected tests:
```
@group exclude-chrome
@group exclude-edge
@group exclude-htmlunit
```
* in the test body:
```
self::skipForUnmatchedBrowsers(['firefox']);
```
* in the test body for the Not Interactable test
```
self::skipForJsonWireProtocol();
```
(Legacy selenium gives a different exception but does not suffer the geckodriver bug).

There is a decrease in coverage in combinations where these tests are excluded.

whimboo:
> The upstream issue is [mozilla/geckodriver#653](https://github.com/mozilla/geckodriver/issues/653) and the gecko issue is https://bugzilla.mozilla.org/show_bug.cgi?id=1374283. I have considered fixing this upstream but it's a crazy complicated problem to solve, I don't currently have time to dedicate to it, and @whimboo is keen to fix this in the spec before fixing it in geckodriver/marionette.

Based on comment 20 on the bug in Bugzilla we would have to create some more test cases including those for multi-line links. Based on that we could then figure out what to do in respect of possible WebDriver spec changes. I'm sorry that this is causing trouble but we still have a limited amount of time for that.

andrewnicols:
@OndraM Not sure why it's still showing as "1 change requested". I've made all of the changes requested and can't find any way to clear the flag.

OndraM (approved):

tests/functional/WebDriverTestCase.php
```diff
@@ -123,6 +123,25 @@ public static function skipForJsonWireProtocol($message = 'Not supported by Json
         }
     }
 
+    /**
+     * Mark a test as skipped if the current browser is not in the list of browsers.
+     *
+     * @param array $browsers List of browsers for this test
```
williamdes, line 129:
```suggestion
     * @param string[] $browsers List of browsers for this test
```

williamdes (approved):

williamdes (approved):

PR #856, opened 2021-01-14

sanduluca:
unescape slashes when encode params

### DIFF
--- a/lib/Remote/HttpCommandExecutor.php
+++ b/lib/Remote/HttpCommandExecutor.php
@@ -326,7 +326,7 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
                 $url
             );
             if (is_array($params) && !empty($params)) {
-                $msg .= sprintf(' with params: %s', json_encode($params));
+                $msg .= sprintf(' with params: %s', json_encode($params, JSON_UNESCAPED_SLASHES));
             }
 
             throw new WebDriverCurlException($msg . "\n\n" . $error);


sanduluca:
Don't quite understand why is it not passing 🤔

OndraM:
Hi, thanks for the contribution! 

Could you please provide some use case why and in what scenario is this change necessary?

(The CI error is probably unrelated, I will have a look.)

sanduluca:
A very common message is one like this
`Curl error thrown for http POST to /session/2104fc08a98b495b6d9d22be647fc3a0/url with params: {"url":"http:\/\/www.example.com\/very\/very\/very\/long\/url"}`
where every slash from URL is escaped and sometimes is hard to understand the link.
I don't  really know why someone will want escaped slashes.
I see this as extra characters, especially when somebody is logging it.

williamdes (approved):

sanduluca:
Here is comment why JSON_UNESCAPED_SLASHES should be treated as default, but not as option
https://stackoverflow.com/a/45681613

OndraM:
@andrewnicols You think this should go to 2.0? I'd say it should be ok for 1.x.

OndraM (approved):

OndraM:
Ping @andrewnicols - do you see any BC break?

OndraM:
Hopefully there is not :slightly_smiling_face: .

OndraM:
Thanks @sanduluca!

PR #861, opened 2021-02-06
Run unit tests via github actions

OndraM:
Travis will stop working quite soon, as Travis gave us free OSS credits only on one-time basis and stopped responding... And the credits are running short.

Functional tests will be the next step, I'm already working on it as well :rocket: .


.github/workflows/php.yaml
```diff
@@ -36,3 +36,57 @@ jobs:
...
+      - name: Run tests
+        run: vendor/bin/phpunit --testsuite unit --colors=always --coverage-clover ./logs/clover.xml
+
+      - name: Submit coverage to Coveralls
+        env:
+          COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COVERALLS_PARALLEL: true
+          COVERALLS_FLAG_NAME: ${{ github.job }}-PHP-${{ matrix.php-version }} ${{ matrix.dependencies }}
+        run: |
+          composer global require php-coveralls/php-coveralls
```
williamdes, line 80: why not using the action for that ?

.github/workflows/php.yaml
```diff
@@ -36,3 +36,57 @@ jobs:
...
+
+      - name: Install dependencies
+        run: composer update --no-interaction ${{ matrix.dependencies }}
+
+      - name: Apply PHPUnit pathes
+        if: ${{ matrix.php-version < 7.2 }}
+        run: scripts/apply-phpunit-patches.sh
+
+      - name: Run tests
+        run: vendor/bin/phpunit --testsuite unit --colors=always --coverage-clover ./logs/clover.xml
```
williamdes, line 72:
```suggestion
        run: composer run test -- --testsuite unit --coverage-clover ./logs/clover.xml
```

.github/workflows/sauce-labs.yaml
```diff
@@ -58,7 +60,7 @@ jobs:
           if [ "$BROWSER_NAME" = "firefox" ]; then EXCLUDE_GROUP+="exclude-firefox,"; fi
           if [ "$BROWSER_NAME" = "chrome" ]; then EXCLUDE_GROUP+="exclude-chrome,"; fi
           if [ -n "$EXCLUDE_GROUP" ]; then EXTRA_PARAMS+=" --exclude-group $EXCLUDE_GROUP"; fi
-          ./vendor/bin/phpunit --coverage-clover ./logs/coverage-clover.xml $EXTRA_PARAMS
+          ./vendor/bin/phpunit --testsuite functional --coverage-clover ./logs/coverage-clover.xml $EXTRA_PARAMS
```
williamdes, line 63: I will not recommend using `composer run test` because the composer timeout would break the running command

williamdes (approved):
Perfect, let's make Travis a tool of the past :+1:

.github/workflows/php.yaml
OndraM: I'm not sure the official coveralls action supports clover reports: https://github.com/coverallsapp/github-action/issues/15

Do you use it anywhere?

.github/workflows/php.yaml
OndraM: Is there some benefit in running it like this, via composer?

.github/workflows/php.yaml
williamdes: Good point, maybe phpunit would have another report type that would be compatible
Maybe open an issue so they can know we need support for clover files

williamdes: And we use the action to mark as finished the uploads
what gives some confusion about why we do not use it here also 
An issue number as a comment would be perfect :)

.github/workflows/php.yaml
williamdes: I would say to be consistent with already defined commands
If you use the command on your dev setup then also use it on the CI to not have different results one day :p

.github/workflows/php.yaml
OndraM: Well the https://github.com/coverallsapp/github-action/issues/15 is exactly about this (clover support). I can add the reference there.

The action for "mark as finished" is used because I wasn't able to make it working via curl, [as mentioned here](https://docs.coveralls.io/parallel-build-webhook), and the action just works.

.github/workflows/php.yaml
OndraM: I see, but I look on the composer commands as "shortcuts" on syntax sugar for developers. Here we use specific settings for CI, so we want kind of "different" result (meaning different output for example).

Also as you mention below, it won't be good for long running tests, so maybe using directly the phpunit command everywhere on CI will be more consistent, in the end :)

williamdes: I agree with the consistency point ^^

Maybe I said this also because my way of learning is by reading code and not manuals, and having a CI check and use dev commands seemed better :)

williamdes (approved):
:100:

williamdes:
Selenium tests are so slow this is the only thing I could not remove from the phpMyAdmin TravisCI

Just thinking about: what about having some website to have a pool of runners that could catch requests to run workflows
Like a pool of volunteer servers
Sorry if this is off-topic but I wanted to have some advice about that idea

Again a free credits for open source subject ^^

OndraM:
> Selenium tests are so slow this is the only thing I could not remove from the phpMyAdmin TravisCI

A whats wrong with GitHub Actions? I think the main issue is usually using too many Selenium tests for things which don't need to be tested via UI. And if its a bigger website with actually many relevant scenarios, you can split tests into multiple builds or use some parallel test runner.

williamdes:
> > Selenium tests are so slow this is the only thing I could not remove from the phpMyAdmin TravisCI
> 
> A whats wrong with GitHub Actions? I think the main issue is usually using too many Selenium tests for things which don't need to be tested via UI. And if its a bigger website with actually many relevant scenarios, you can split tests into multiple builds or use some parallel test runner.

We have a lot of tests and each selenium run uses 30 minutes of build times
But yeah there is some work to be done
I would prefer not removing the tests because they also test our JS code that for now has no unit tests

OndraM:
> We have a lot of tests and each selenium run uses 30 minutes of build times

But GH Actions should have time [long enough limits](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration):
> Each job in a workflow can run for up to 6 hours of execution time.

williamdes:
> > We have a lot of tests and each selenium run uses 30 minutes of build times
> 
> But GH Actions should have time [long enough limits](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration):
> 
> > Each job in a workflow can run for up to 6 hours of execution time.

Limits are not what worries me, it is the credits spent each minute

OndraM:
> Limits are not what worries me, it is the credits spent each minute

But GH Actions are free for public repositories, so there is no need to worry, isn't it?

williamdes:
> > Limits are not what worries me, it is the credits spent each minute
> 
> But GH Actions are free for public repositories, so there is no need to worry, isn't it?

You can find how much you have per month here : 
https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions

That said it seems true 
For me the documentation was unclear that it only counts private repositories

This is very good news as I will be able to remove TravisCI completely 

My organization settings seem to agree with you
Screenshot_20210208_120854_com brave browser

PR #863, opened 2021-02-11
Run functional tests on GitHub Actions, remove Travis

OndraM:
Well, that wasn't that simple 🙃. But as our Travis is now disabled, I want to do this before doing any new release, to be sure tests are passing.

Implementation note: We cannot use Selenium docker images, as we need to control which browser is actually being started (and there eg. isn't `if` option for GH Actions service containers). **For projects running selenium tests using this library this is not necessary - using service containers would be much easier.** (See [example in phpmyadmin tests](https://github.com/williamdes/phpmyadmintest/blob/fda5195be7dc84ad5c2a76d0899e11efed06766f/.github/workflows/test-selenium.yml#L28-L38) by @williamdes ).

However here we need to run everything natively right on the OS. Thankfully, Chrome, ChromeDriver, Firefox and GeckoDriver are [by default installed on GitHub Actions Ubuntu image](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#browsers-and-drivers), so we at least don't have to worry with downloading the latest & compatible version.

As a result, we have all the builds like on Travis and more (like total running time ca. ~4x faster :racing_car: ), except:
- HtmlUnit (no longer part of selenium jar, also headless chrome is also fast and supports javascript completely; but we keep on ChromeDriver build with disabled W3C protocol to make sure we don't break OSS protocol support). But we also have separate unit-tests build.
- Legacy Firefox (version 45 and older). Its over 5 years old, and there is no benefit in trying to run test there. Hopefully everyone already moved to either GeckoDriver or Chrome.

(And as a coincidence - thanks to different run environment, I also found and fixed bug in #835).

Suggestions and simplifications welcomed!


.github/workflows/php.yaml
```diff
@@ -82,9 +82,108 @@ jobs:
...
+        env:
+          COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COVERALLS_PARALLEL: true
+          COVERALLS_FLAG_NAME: ${{ github.job }}-${{ matrix.browser }}-selenium-${{ matrix.selenium-server }}-w3c-${{ matrix.w3c }}
+        run: |
+          composer global require php-coveralls/php-coveralls
+          ~/.composer/vendor/bin/php-coveralls -v
+
+      - name: Print logs
+        if: ${{ always() }}
```
williamdes, line 177: I used success or failure because it seemed like cancel would run this step

lib/Remote/Service/DriverService.php
```diff
@@ -110,25 +110,24 @@ public function isRunning()
...
 
     /**
-     * Check if the executable is executable.
-     *
      * @param string $executable
      * @throws Exception
-     * @return string
      */
-    protected static function checkExecutable($executable)
+    protected function setExecutable($executable)
```
williamdes, line 116: Bc break?

williamdes (approved):
Good job!

.github/workflows/php.yaml
OndraM: But if the build is cancelled by github because other build failed, you may still want this build to dump its logs, no?

lib/Remote/Service/DriverService.php
OndraM: 😕

OndraM: I added the `protected static function checkExecutable($executable)` back. But it just returns the value. However the API is kept, and executable is now checked when calling setExectauble from the constructor.

williamdes (approved):

.github/workflows/php.yaml
williamdes: It would make sense, that said I was thinking that if you implement `[ci skip]` like I did for phpMyAdmin it could run and make no sense

PR #866, opened 2021-02-16
Remove FF47 saucelabs build

OndraM:
Exclusions and conditions in tests for the old FF were already removed in #863 , so now the old-FF tests on saucelabs are failing.


OndraM:
Removed in favor of #867 (builds from forks cannot run saucelabs).

PR #867, opened 2021-02-16
Remove Firefox 47 saucelabs build

OndraM:
Exclusions and conditions in tests for the old FF were already removed in #863 , so now the old-FF tests on saucelabs are failing.

### DIFF
--- a/.github/workflows/sauce-labs.yaml
+++ b/.github/workflows/sauce-labs.yaml
@@ -16,12 +16,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-#         # Firefox 47 is the last Firefox with OSS protocol support
-          - { name: "Firefox 47, OSS protocol", BROWSER_NAME: "firefox", VERSION: "47.0", PLATFORM: "Windows 10", DISABLE_W3C_PROTOCOL: "1", tunnel-id: "gh-1" }
           # Chrome 74 is the last version which doesn't use W3C WebDriver by default and rather use OSS protocol
-          - { name: "Chrome 74, OSS protocol", BROWSER_NAME: "chrome", VERSION: "74.0", PLATFORM: "Windows 10", DISABLE_W3C_PROTOCOL: "1", tunnel-id: "gh-2" }
-          - { name: "Chrome latest, W3C protocol", BROWSER_NAME: "chrome", VERSION: "latest", PLATFORM: "Windows 10", tunnel-id: "gh-3" }
-          - { name: "Edge latest, W3C protocol", BROWSER_NAME: "MicrosoftEdge", VERSION: "latest", PLATFORM: "Windows 10", tunnel-id: "gh-4" }
+          - { name: "Chrome 74, OSS protocol", BROWSER_NAME: "chrome", VERSION: "74.0", PLATFORM: "Windows 10", DISABLE_W3C_PROTOCOL: "1", tunnel-id: "gh-1" }
+          - { name: "Chrome latest, W3C protocol", BROWSER_NAME: "chrome", VERSION: "latest", PLATFORM: "Windows 10", tunnel-id: "gh-2" }
+          - { name: "Edge latest, W3C protocol", BROWSER_NAME: "MicrosoftEdge", VERSION: "latest", PLATFORM: "Windows 10", tunnel-id: "gh-3" }
 
     name: ${{ matrix.name }} (${{ matrix.tunnel-id }})
     steps:
--- a/.github/workflows/sauce-labs.yaml
+++ b/.github/workflows/sauce-labs.yaml
@@ -15,9 +15,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        w3c: [true]
         include:
           # Chrome 74 is the last version which doesn't use W3C WebDriver by default and rather use OSS protocol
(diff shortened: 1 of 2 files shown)


williamdes (approved):
We do not test anymore FF ?

OndraM:
> We do not test anymore FF ?

We still do test against latest FF in integration tests. But this is removing Saucelabs build against the old FF 47. There is still OSS protocol build against saucelabs (via Chrome).

williamdes:
> > We do not test anymore FF ?
> 
> We still do test against latest FF in integration tests. But this is removing Saucelabs build against the old FF 47. There is also still OSS protocol build against saucelabs (via Chrome).

But could you replace it with the Latest FF ?
Would that make sense ?

andrewnicols:
Yeah, it would be good to test against latest Firefox. Maybe the tests can use a docker image instead of Saucelabs in future too? Would be faster than remote browser testing.

.github/workflows/sauce-labs.yaml
```diff
@@ -15,13 +15,12 @@ jobs:
...
-          - { name: "Firefox 47, OSS protocol", BROWSER_NAME: "firefox", VERSION: "47.0", PLATFORM: "Windows 10", DISABLE_W3C_PROTOCOL: "1", tunnel-id: "gh-1" }
           # Chrome 74 is the last version which doesn't use W3C WebDriver by default and rather use OSS protocol
-          - { name: "Chrome 74, OSS protocol", BROWSER_NAME: "chrome", VERSION: "74.0", PLATFORM: "Windows 10", DISABLE_W3C_PROTOCOL: "1", tunnel-id: "gh-2" }
-          - { name: "Chrome latest, W3C protocol", BROWSER_NAME: "chrome", VERSION: "latest", PLATFORM: "Windows 10", tunnel-id: "gh-3" }
-          - { name: "Edge latest, W3C protocol", BROWSER_NAME: "MicrosoftEdge", VERSION: "latest", PLATFORM: "Windows 10", tunnel-id: "gh-4" }
+          - { name: "Chrome 74, OSS protocol", BROWSER_NAME: "chrome", VERSION: "74.0", PLATFORM: "Windows 10", w3c: false, tunnel-id: "gh-1" }
+          - { name: "Chrome latest, W3C protocol", BROWSER_NAME: "chrome", VERSION: "latest", PLATFORM: "Windows 10", tunnel-id: "gh-2" }
+          - { name: "Edge latest, W3C protocol", BROWSER_NAME: "MicrosoftEdge", VERSION: "latest", PLATFORM: "Windows 10", tunnel-id: "gh-3" }
```
andrewnicols, line 23: Would be good t have a Firefox latest version here.

.github/workflows/tests.yaml
```diff
@@ -108,7 +107,6 @@ jobs:
           php-version: 7.4
           extensions: mbstring, intl, zip
           coverage: xdebug
-          tools: composer:v2
```
andrewnicols, line 111: Is this now the default or is this an unrelated change?

andrewnicols (changes requested):

.github/workflows/tests.yaml
williamdes: Maybe a copy paste from my example :smile: 
I forced it on phpMyAdmin to have more performance and have less time spent installing the composer modules

andrewnicols: But that's a removal of v2 - so slower.

OndraM:
> Yeah, it would be good to test against latest Firefox. Maybe the tests can use a docker image instead of Saucelabs in future too? Would be faster than remote browser testing.

But we still test against latest firefox, there are two build in functional tests:
- https://github.com/php-webdriver/php-webdriver/runs/1910337096?check_suite_focus=true (via Selenium server)
- https://github.com/php-webdriver/php-webdriver/runs/1910337115?check_suite_focus=true (only with geckodriver)

This is latest Firefox is installed directly in Ubuntu VM used by GH Actions, so this is probalby even faster then in docker and we are closer to real Firefox deployment than Firefox hidden in Saucelabs.

The change here is we don't test against Firefox via Saucelabs. I don't see there real value: Saucelabs is mainy for browsers/versions, we cannot run locally (which is always faster and less flaky). This is now mainly MS Edge on Windows. The two other saucelab builds are Chrome on Windows with OSS and Chrome on Windows with W3C protocol, but I take them as a sanity check - to make sure we don't break anything for people running tests against Saucelabs (as they have their custom implementation of the remote selenium endpoint, with different behavior than Selenium server). 

I would be really considerate with any other saucelabs build - they are much slower and less reliable. And especially if we can achieve the same locally or in docker, it should be preferred. And we already have latest Firefox build. What would be a value of such additional build with the same browser, would it be worth the cost and effort?

.github/workflows/tests.yaml
OndraM: >  The latest stable version of composer is set up by default.
https://github.com/marketplace/actions/setup-php-action#wrench-tools-support

And Composer 2 is in the stable channel for a while. See build output:
https://github.com/php-webdriver/php-webdriver/runs/1910336904?check_suite_focus=true#step:3:12

andrewnicols:
In which case +1 :)

andrewnicols (approved):

williamdes (approved):

andrewnicols:
@OndraM , do you need any more from us?

OndraM:
> @OndraM , do you need any more from us?

No, only a time :smile: . Merging now :rocket:

Issue #345, opened 2016-10-16
Support local safaridriver from Safari 10.0


OndraM:
Safari could be started through Selenium Server as remote WebDriver.

Since Safari 10.0 ([see blogpost](https://webkit.org/blog/6900/webdriver-support-in-safari-10/)), however, it could also be started also locally, using binary "safaridriver" (as chrome, opera and geckodriver - see also #344, #342). This is already supported by other selenium language bindings, so we could support it as well.  

It would require creating new classes SafariDriver, SafariDriverService and SafariOptions - see [java bindings](https://github.com/SeleniumHQ/selenium/tree/c2a4dabfdf0a29a252c9dd33aa541c06c061a9d4/java/client/src/org/openqa/selenium/safari) as an example.

rajabhaskar:
@OndraM Can you please let me know how I can run acceptance suite in safari browser in Windows 8.1.  Below is my acceptance.yml file configuration.  I tried --env safari and --env firefox by mentioning the env in acceptance suite. yml also but no luck. it's giving me error 

_**  [Codeception\Exception\ConnectionException]
  Curl error thrown for http POST to /session with params: {"desiredCapabilit
  ies":{"unexpectedAlertBehaviour":"accept","handlesAlerts":true,"browserName
  ":"firefox"}}
  Could not resolve host: selenium-hub

  Please make sure that Selenium Server or PhantomJS is running.**_

But I do have selenium jar or phantom running aside already with below command
_**phantomjs --webdriver=4444  **_  [OR] _**java -jar vendor/selenium/selenium-server/selenium-server-standalone-2.53.0.jar**_

**Acceprance.suite.yml**
modules:
enabled:
- WebDriver
- \Helper\Acceptance
- \Helper\Locator
- \Helper\User
- \Helper\TestData
config:
WebDriver:
url: 'https://www.abc.com'
browser: safari
wait: 2
window_size: maximize
clear_cookies: false
capabilities:
unexpectedAlertBehaviour: 'accept'
handlesAlerts: true

but it is not working. If I replace the browser name with Chrome (and run executable) or firefox then I was able to run my suite. Please advise. (By the way, I do have safari installed manually in my machine - V5.1.7)
Thanks

AntonRosen:
Hi @OndraM,
I'm working with php via linux command line and trying to run a few tests. I've read all of your posts related to this and the links you provided, but I can't find anything that helps with my problem (or maybe I just didn't understand the solution when I saw it)

[...]

PR #874, opened 2021-03-18
Add Safari on macos-latest build

oleg-andreyev:


williamdes (approved):

OndraM:
Hi @oleg-andreyev , great job working on this! :medal_sports: 

One of my thoughts regarding this is to split it into two pull requests, as different issues may arise from them and would be safer to  to take care of them separately:

1) Executing current tests using safaridriver executable on macos. This may need some fixes in the library, or more probably some Safaridriver issues will arise, so some tests would need to be skipped. In such case, if you find related bugreport from Safari, it would be great to link it in tests which need to be skipped in safari. You may also need to add "exclude-safari" condition into `tests.yaml`. For this first part, the `SafariDriver.php` implementation is not needed, this my suggestion for separating this.

2) Implementing `SafariDriver` (which is reported as #345) and adding a tests for it. See `ChromeDriverTest` for reference - the new test would obviously have to be limited to macos (using [`@require` annotation](https://phpunit.readthedocs.io/en/9.5/incomplete-and-skipped-tests.html#incomplete-and-skipped-tests-requires-tables-api)) . 

Let me know if I can provide any help or guidance!

lib/Safari/SafariDriverService.php
```diff
@@ -0,0 +1,37 @@
...
+
+use Facebook\WebDriver\Remote\Service\DriverService;
+
+class SafariDriverService extends DriverService
+{
+    /**
+     * The environment variable storing the path to the chrome driver executable.
+     * @deprecated Use ChromeDriverService::CHROME_DRIVER_EXECUTABLE
+     */
+    const SAFARI_DRIVER_EXE_PROPERTY = 'webdriver.safari.driver';
```
OndraM, line 13: We don't need to introduce this at all

lib/Safari/SafariDriver.php
```diff
@@ -0,0 +1,98 @@
...
+
+    public function startSession(DesiredCapabilities $desired_capabilities)
+    {
+        $command = new WebDriverCommand(
+            null,
+            DriverCommand::NEW_SESSION,
+            [
+                'capabilities' => [
+                    'firstMatch' => [(object) $desired_capabilities->toW3cCompatibleArray()],
+                ],
```
OndraM, line 41: Does SafariDriver even support OSS protocol? Maybe we can only support the W3C one?

.github/workflows/tests.yaml
```diff
@@ -82,19 +82,21 @@ jobs:
...
       matrix:
+        os: ['ubuntu-latest']
         browser: ['chrome', 'firefox']
         selenium-server: [true, false] # Whether to run via Selenium server or directly via browser driver
         w3c: [true] # Although all builds negotiate protocol by default, it implies W3C protocol for both Chromedriver and Geckodriver
         include:
           # Force OSS (JsonWire) protocol on ChromeDriver - to make sure we keep compatibility
-          - { browser: 'chrome', selenium-server: false, w3c: false }
+          - { browser: 'chrome', os: 'ubuntu-latest', selenium-server: false, w3c: false }
+          - { browser: 'safari', os: 'macos-latest', selenium-server: false }
```
OndraM, line 99: w3c must be implicitly set to true, otherwise strange things will happen (the scripts expects the value to be set to true or false)

oleg-andreyev: Based on https://developer.apple.com/documentation/webkit/macos_webdriver_commands_for_safari_12_and_later, since Safari 12, it's only W3C

lib/Safari/SafariDriverService.php
oleg-andreyev: removed.

lib/Safari/SafariDriver.php
oleg-andreyev: Actually Safari < 12 support OSS and GitHub Actions support macOS-11,
Safari >= 12 support only W3C and it's GitHub Actions support macOS-latest

oleg-andreyev:
@OndraM I've marked macOS build as experimental, so we can merge this as is and continue in other PRs.

lib/Safari/SafariDriverService.php
```diff
@@ -0,0 +1,33 @@
...
+     * @return static
+     */
+    public static function createDefaultService()
+    {
+        $pathToExecutable = getenv(self::SAFARI_DRIVER_EXECUTABLE);
+        if ($pathToExecutable === false) {
+            $pathToExecutable = static::DEFAULT_EXECUTABLE;
+        }
+
+        $port = 9515; // TODO: Get another port if the default port is used.
```
williamdes, line 28:
```suggestion
        $port = getenv(self::SAFARI_DRIVER_PORT);
        if ($port === false) {
            $port = static::DEFAULT_PORT;
        }
```

williamdes (approved):

.github/workflows/tests.yaml
```diff
@@ -135,18 +141,25 @@ jobs:
...
           geckodriver --version
           xvfb-run --server-args="-screen 0, 1280x720x24" --auto-servernum \
-          geckodriver &> ./logs/geckodriver.log &
+          geckodriver --port=4444 &> ./logs/geckodriver.log &
+
+      - name: Start SafariDriver
+        if: ${{ !matrix.selenium-server && matrix.browser == 'safari' }}
+        run: |
+          defaults read /Applications/Safari.app/Contents/Info CFBundleShortVersionString
+          /usr/bin/safaridriver -p 4444 --diagnose &
```
williamdes, line 150: according to your code the driver port if 7055
did I misunderstand ?

williamdes (approved):

.github/workflows/tests.yaml
```diff
@@ -178,6 +191,9 @@ jobs:
           if [ -f ./logs/chromedriver.log ]; then cat ./logs/chromedriver.log; fi
           if [ -f ./logs/geckodriver.log ]; then cat ./logs/geckodriver.log; fi
           if [ -f ./logs/php-server.log ]; then cat ./logs/php-server.log; fi
+          if [ "$BROWSER_NAME" = "safari" ]; then
```
OndraM, line 194: Why not use just `-f` as with the other logs?

.github/workflows/tests.yaml
```diff
@@ -135,18 +141,25 @@ jobs:
...
-          php -S 127.0.0.1:8000 -t tests/functional/web/ &>>./logs/php-server.log &
+          php -S 127.0.0.1:8000 -t tests/functional/web/ &> ./logs/php-server.log &
 
       - name: Wait for browser & PHP to start
         run: |
           while ! nc -z localhost 4444 </dev/null; do echo Waiting for remote driver to start...; sleep 1; done
           while ! nc -z localhost 8000 </dev/null; do echo Waiting for PHP server to start...; sleep 1; done
 
       - name: Run tests
+        continue-on-error: ${{ matrix.experimental }}
```
OndraM, line 162: Isn't it redundant to have this on both job and step level?

.github/workflows/tests.yaml
```diff
@@ -82,21 +82,27 @@ jobs:
...
+        os: ['ubuntu-latest']
         browser: ['chrome', 'firefox']
+        experimental: [false]
         selenium-server: [true, false] # Whether to run via Selenium server or directly via browser driver
         w3c: [true] # Although all builds negotiate protocol by default, it implies W3C protocol for both Chromedriver and Geckodriver
         include:
           # Force OSS (JsonWire) protocol on ChromeDriver - to make sure we keep compatibility
-          - { browser: 'chrome', selenium-server: false, w3c: false }
+          - { browser: 'chrome', os: 'ubuntu-latest', selenium-server: false, w3c: false, experimental: false }
+          - { browser: 'safari', os: 'macos-11.0', selenium-server: false, w3c: false, experimental: true }
```
OndraM, line 100: This is now failing on `No runner matching the specified labels was found: macos-11.0`. 

Also I'd prefer to remove this non-W3C build altogether. There is no value in trying to run tests, which we were not running at all until now, on deprecated versions of Safari and with deprecated OSS protocol.

OndraM (changes requested):
Hi, thanks for the updates!

As this is moving forward, the previous suggestion seems necessary: so could you please squash the changes and separate it into tow pull requests: keeping in this one only the changes in GH Actions + related tests, and separate the SafariDriver into new pull request, so we don't mix those separated issues things together?

I don't want to merge failing builds (hidden by continue-on-error) now and fix it later: we should add macos build and fix (or skip) the tests as part of the PR (so that the builds are actually passing). 

This does not need to be mixed with the SafariDriver implementation, this is why I'd like to have in its separate pull requests. Also the SafariDriver is now not covered by tests, they should be added before merging this into upstream.

Thank you :)

.github/workflows/tests.yaml
```diff
@@ -82,21 +82,27 @@ jobs:
...
+          - { browser: 'safari', os: 'macos-latest', selenium-server: false, w3c: true, experimental: true }
+
+    continue-on-error: ${{ matrix.experimental }}
```
OndraM, line 103: I think we should try to make the build passing before it is merged, and not add this.

.github/workflows/tests.yaml
oleg-andreyev: I'm going split this PR into two as @OndraM is suggesting, so it can be ignored.
But default port of Safari Driver is somewhat 7055 (nothing is said about it in official docs)

.github/workflows/tests.yaml
oleg-andreyev: Ok, will remove it. Maybe idea was to validate if library even works with OSS SafariDriver.

.github/workflows/tests.yaml
oleg-andreyev: Because I don't know the exact file.

```
❯ ls ~/Library/Logs/com.apple.WebDriver/safaridriver*
/Users/oandreyev/Library/Logs/com.apple.WebDriver/safaridriver.44337.qcntit.txt
/Users/oandreyev/Library/Logs/com.apple.WebDriver/safaridriver.45207.spaixu.txt
/Users/oandreyev/Library/Logs/com.apple.WebDriver/safaridriver.45618.mpzsjj.txt
/Users/oandreyev/Library/Logs/com.apple.WebDriver/safaridriver.46130.rzecqu.txt
/Users/oandreyev/Library/Logs/com.apple.WebDriver/safaridriver.52330.tijxyb.txt
~ ❯    
```

.github/workflows/tests.yaml
oleg-andreyev: My initial idea was to add macOS as experimental (allow to fail) and try to fix tests in next PR.
To avoid spending too much time on this PR and allow others to contribute.

oleg-andreyev: And here is one example:
https://feedbackassistant.apple.com/feedback/9051272

Safaridriver does not persist "pressed" state of Shift

OndraM: The URL requires Apple ID to login...

When there are some tests which do not pass on some browser, we mark them explicitly with `@group exclude-[browser]`, in this case it will be `@group exclude-safari`.

 We have such cases for other browsers as well (chrome, firefox, edge), [for example here](https://github.com/php-webdriver/php-webdriver/blob/a82056dfe7bb1dc3c1d912501f55cd06da5e3cf5/tests/functional/RemoteKeyboardTest.php#L13-L17). Ideally, there should be also a link to some safaridriver bug tracking, if possible.

Also `exclude-safari` should be added [below](https://github.com/php-webdriver/php-webdriver/blob/main/.github/workflows/tests.yaml#L158-L159) to the GH Actions configuration.

When we mark all problematic tests by the group, and thanks to this, it will cause tests not working on safari to be skipped (so they won't fail). So it will allow as to not use the `continue-on-error` at all, am I right? Because the safari tests will be working as the others do.

.github/workflows/tests.yaml
```diff
@@ -127,19 +132,25 @@ jobs:
...
+          chromedriver --port=4444 &> ./logs/chromedriver.log &
 
       - name: Start GeckoDriver
         if: ${{ !matrix.selenium-server && matrix.browser == 'firefox' }}
         run: |
           firefox --version
           geckodriver --version
           xvfb-run --server-args="-screen 0, 1280x720x24" --auto-servernum \
-          geckodriver &> ./logs/geckodriver.log &
+          geckodriver --port=4444 &> ./logs/geckodriver.log &
```
OndraM, line 143: Why this change? The port `4444` is default

.github/workflows/tests.yaml
```diff
@@ -127,19 +132,25 @@ jobs:
         run: |
           google-chrome --version
           xvfb-run --server-args="-screen 0, 1280x720x24" --auto-servernum \
-          chromedriver --port=4444 --url-base=/wd/hub &> ./logs/chromedriver.log &
```
OndraM, line 130: Currently this path is by default expected [in the WebDriverTestCase](https://github.com/php-webdriver/php-webdriver/blob/main/tests/functional/WebDriverTestCase.php#L24).

This is causing chrome test to fail: https://github.com/php-webdriver/php-webdriver/pull/874/checks?check_run_id=2170530712

So I guess this should be kept.

OndraM (changes requested):

OndraM:
@oleg-andreyev Hi there, can I provide some help in moving this forward?

oleg-andreyev:
@OndraM having hard times to find time to work on this. Won't promise anything but I'll try to finish this work on this weekends.

.github/workflows/tests.yaml
oleg-andreyev: reverted

.github/workflows/tests.yaml
oleg-andreyev: reverted

OndraM:
@oleg-andreyev Thanks, let me know if I can provide aby assistance.

Also if you will be working on the following pull request with `SafariDriver`, `SafariDriverService` etc., first have a look at new [`FirefoxDriver`](https://github.com/php-webdriver/php-webdriver/blob/main/lib/Firefox/FirefoxDriver.php) and [`FirefoxDriverService`](https://github.com/php-webdriver/php-webdriver/blob/main/lib/Firefox/FirefoxDriverService.php). They were added in the meantime, and may serve as an example how the SafariDriver could be now integrated (also using new [`LocalWebDriver`](https://github.com/php-webdriver/php-webdriver/blob/main/lib/Local/LocalWebDriver.php)). You can also push your previous SafariDriver changes to some branch and I can continue the work - but of course it is preferred if you'd be able to submit the change.

Thanks again :+1:

tests/functional/RemoteWebDriverCreateTest.php
```diff
@@ -40,7 +42,12 @@ public function testShouldStartBrowserAndCreateInstanceOfRemoteWebDriver()
 
         // MicrosoftEdge on Sauce Labs started to identify itself back as "msedge"
         if ($this->desiredCapabilities->getBrowserName() !== WebDriverBrowserType::MICROSOFT_EDGE) {
-            $this->assertSame($this->desiredCapabilities->getBrowserName(), $returnedCapabilities->getBrowserName());
+            if ($this->desiredCapabilities->getBrowserName() === WebDriverBrowserType::SAFARI) {
+                // TODO: ???
```
oleg-andreyev, line 46: safaridriver is returning `Safari` (Capital) vs `safari` expected

@OndraM what's the purpose of this assert?

tests/functional/RemoteWebElementTest.php
```diff
@@ -23,6 +23,10 @@ public function testShouldGetText()
         $elementWithTextWithSpaces = $this->driver->findElement(WebDriverBy::id('text-with-spaces'));
 
         $this->assertEquals('Foo bar text', $elementWithSimpleText->getText());
+
+        // TODO: Safari is returning text with newline? Should getText manipulate with text?
```
oleg-andreyev, line 27: @OndraM from specification it's unclear should getText return either "newline" or not. In specifications it's said "as rendered", so from this perspective, I should be getting "newline" and spaces between, so probably chromedriver and geckodriver misbehave.

https://www.w3.org/TR/webdriver/#get-element-text

tests/functional/RemoteWebElementTest.php
```diff
@@ -422,8 +426,13 @@ public function testShouldTakeAndSaveElementScreenshot()
 
         // Assert file output
         $imageFromFile = imagecreatefrompng($screenshotPath);
-        $this->assertEquals(5, imagesx($imageFromFile));
-        $this->assertEquals(5, imagesy($imageFromFile));
+        if (getenv('BROWSER_NAME') === 'safari') {
```
oleg-andreyev, line 429: On hidpi (retina): 5px => 10px, not sure how to detect it...

tests/functional/WebDriverTestCase.php
```diff
@@ -67,32 +67,30 @@ protected function setUp(): void
...
-                null
-            );
-        }
+        $this->createWebDriver();
     }
 
     protected function tearDown(): void
     {
-        if ($this->driver instanceof RemoteWebDriver && $this->driver->getCommandExecutor()) {
+        if ($this->driver) {
```
OndraM, line 82: We should compare `$this->driver` with something, instead of just loose truthy comparison...

.github/workflows/tests.yaml
```diff
@@ -82,21 +82,26 @@ jobs:
...
+        experimental: [false]
         selenium-server: [true, false] # Whether to run via Selenium server or directly via browser driver
         w3c: [true] # Although all builds negotiate protocol by default, it implies W3C protocol for both Chromedriver and Geckodriver
         include:
           # Force OSS (JsonWire) protocol on ChromeDriver - to make sure we keep compatibility
-          - { browser: 'chrome', selenium-server: false, w3c: false }
+          - { browser: 'chrome', os: 'ubuntu-latest', selenium-server: false, w3c: false, experimental: false }
+          - { browser: 'safari', os: 'macos-latest', selenium-server: false, w3c: true, experimental: true }
+
+    continue-on-error: ${{ matrix.experimental }}
```
OndraM, line 102: I'm still a bit confused why this is needed. The test should be passing with all the exclude-group, right? So why then is this "experimental" flag needed?

OndraM (changes requested):
Nice work @oleg-andreyev :+1: 

Just a few questions and suggestions below:

tests/functional/WebDriverTestCase.php
oleg-andreyev: added `!== null`

tests/functional/RemoteWebElementTest.php
oleg-andreyev: it's already skipped, but I'll leave a to do comment

tests/functional/RemoteWebDriverCreateTest.php
oleg-andreyev: used `assertEqualsIgnoringCase`

.github/workflows/tests.yaml
oleg-andreyev: removed

tests/functional/Firefox/FirefoxDriverTest.php
```diff
@@ -10,6 +10,7 @@
 
 /**
  * @group exclude-saucelabs
+ * @group exclude-safari
```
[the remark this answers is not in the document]
oleg-andreyev: removed

lib/Remote/RemoteWebElement.php
```diff
@@ -204,10 +204,21 @@ public function getLocation()
...
     public function getLocationOnScreenOnceScrolledIntoView()
     {
-        $location = $this->executor->execute(
-            DriverCommand::GET_ELEMENT_LOCATION_ONCE_SCROLLED_INTO_VIEW,
-            [':id' => $this->id]
-        );
+        if ($this->isW3cCompliant) {
+            $result = $this->executor->execute(DriverCommand::EXECUTE_SCRIPT, [
+                'script' => sprintf(
+                    'return %s.getBoundingClientRect();',
```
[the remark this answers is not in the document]
oleg-andreyev: done

tests/functional/Chrome/ChromeDevToolsDriverTest.php
```diff
@@ -7,8 +7,6 @@
 
 /**
  * @group exclude-saucelabs
- * @group exclude-firefox
- * @group exclude-edge
```
oleg-andreyev, line 11: based on setUp method

OndraM: This unfortunately must be kept for any skipped tests - otherwise the tests are initialized on sauce labs but their status is never set finished (because they are lates skipped). And some unfinished tests cause saucelabs dashboard to report the whole build as failed... So we must not even star the tests.

williamdes (approved):

PR #890, opened 2021-04-16
Implement FirefoxOption class

OndraM:
This adds `FirefoxOptions`, which - similarly to [ChromeOptions](https://github.com/php-webdriver/php-webdriver/wiki/Chrome#chromeoptions) - provides language-level data object to construct value of `moz:firefoxOptions`. Until now, this was possible only using multidiemensional arrays.

One nice feature is the possibility to directly pass `about:config` preferences to this object.

This also changes the way how default preferences (which we add to Firefox) are passed to the Capabilities. It is no longer via configured FirefoxProfile, but rather directly using `prefs` key in firefoxOptions, which is the recommended way. (Because of this, there is a BC layer in DesiredCapabilities, otherwise default options will be overwritten for those who already use `moz:firefoxOptions`).

📙 As part of this, I created [Firefox](https://github.com/php-webdriver/php-webdriver/wiki/Firefox) wiki page, which covers all of this and more. Suggestions welcome!

### Example usage
```php
// Configure FirefoxOptions
$firefoxOptions = (new FirefoxOptions())
    ->addArguments(['-headless']); // add Firefox CLI arguments
    ->setPreference('dom.ipc.processCount', 4); // set about:config preference
    ->setOption('binary', '/home/user/Downloads/firefox-esr'); // use this to set directly other firefoxOptions, see https://developer.mozilla.org/en-US/docs/Web/WebDriver/Capabilities/firefoxOptions

// Add it to DesiredCapabilities
$capabilities = DesiredCapabilities::firefox();
$capabilities->setCapability(FirefoxOptions::CAPABILITY, $firefoxOptions);

// Start the browser
$driver = RemoteWebDriver::create($serverUrl, $capabilities);
```

### Previously it was only possible using array
```php
$capabilities = DesiredCapabilities::firefox();
// Configure moz:firefoxOptions using array
$capabilities->setCapability(
    'moz:firefoxOptions',
    [
        'args' => ['-headless'],
        'prefs' => ['dom.ipc.processCount' => 4],
        'binary' => '/home/user/Downloads/firefox-esr',
    ]
);

// Start the browser
$driver = RemoteWebDriver::create($serverUrl, $capabilities);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 This project versioning adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+- Add `FirefoxOptions` class to simplify passing Firefox capabilities. Usage is covered [in our wiki](https://github.com/php-webdriver/php-webdriver/wiki/Firefox#firefoxoptions).
 
 ## 1.10.0 - 2021-02-25
 ### Added
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ Desired capabilities define properties of the browser you are about to start.
 They can be customized:
 
 ```php
+use Facebook\WebDriver\Firefox\FirefoxOptions;
 use Facebook\WebDriver\Remote\DesiredCapabilities;
 
 $desiredCapabilities = DesiredCapabilities::firefox();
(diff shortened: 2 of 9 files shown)


williamdes (approved):
Nice

PR #892, opened 2021-04-20
Implement FirefoxDriver

### LINKED ISSUE #891: Implement FirefoxDriver
To be able to start Firefox without need to running `geckodriver` binary separately, we could implement FirefoxDriver.

This would work similarly as already existing ChromeDriver ([source](https://github.com/php-webdriver/php-webdriver/blob/main/lib/Chrome/ChromeDriver.php), [usage example in wiki](https://github.com/php-webdriver/php-webdriver/wiki/Chrome#start-directly-using-chromedriver-class)).

There may be also useful some code generalization and cleanup in those related classes (ChromeDriver, ChromeDriverService...) to DRY.

Also related to #345 (SafariDriver).

OndraM:
This adds FirefoxDriver. Now you are able to start Firefox locally without need to run `geckodriver` binary separately. The usage is [similar to ChromeDriver class](https://github.com/php-webdriver/php-webdriver/wiki/Chrome#start-directly-using-chromedriver-class).

There is also new `LocalWebDriver`, which is used as a base class for local browser drivers implementations (currently ChromeDriver and now FirefoxDriver).

Fixes #891.

#### Basic example:
```php
$driver = FirefoxDriver::start();
$driver->get('https://google.com/');
```

#### Using new FirefoxOptions from #890:
```php
$firefoxOptions = new FirefoxOptions();
$firefoxOptions->addArguments(['-headless']);

$capabilities = new DesiredCapabilities([FirefoxOptions::CAPABILITY => $firefoxOptions]);

$driver = FirefoxDriver::start($capabilities);
$driver->get('https://google.com/');
```

#### TODO
- [x] Update [wiki](https://github.com/php-webdriver/php-webdriver/wiki/Firefox) to include `FirefoxDriver::start`

### DIFF
--- a/lib/Firefox/FirefoxDriverService.php
+++ b/lib/Firefox/FirefoxDriverService.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Facebook\WebDriver\Firefox;
+
+use Facebook\WebDriver\Remote\Service\DriverService;
+
+class FirefoxDriverService extends DriverService
+{
+    /**
+     * @var string Name of the environment variable storing the path to the driver binary
+     */
+    const WEBDRIVER_FIREFOX_DRIVER = 'WEBDRIVER_FIREFOX_DRIVER';
+    /**
+     * @var string Default executable used when no other is provided
+     * @internal
+     */
+    const DEFAULT_EXECUTABLE = 'geckodriver';
+
+    /**
+     * @return static
+     */
+    public static function createDefaultService()
+    {
+        $pathToExecutable = getenv(static::WEBDRIVER_FIREFOX_DRIVER);
+        if ($pathToExecutable === false || $pathToExecutable === '') {
+            $pathToExecutable = static::DEFAULT_EXECUTABLE;
+        }
+
+        $port = 9515; // TODO: Get another free port if the default port is used.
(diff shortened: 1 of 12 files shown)


lib/Firefox/FirefoxDriverService.php
```diff
@@ -0,0 +1,34 @@
...
+     * @return static
+     */
+    public static function createDefaultService()
+    {
+        $pathToExecutable = getenv(static::WEBDRIVER_FIREFOX_DRIVER);
+        if ($pathToExecutable === false || $pathToExecutable === '') {
+            $pathToExecutable = static::DEFAULT_EXECUTABLE;
+        }
+
+        $port = 9515; // TODO: Get another free port if the default port is used.
```
williamdes, line 29: getenv ?

williamdes (approved):

lib/Firefox/FirefoxDriverService.php
OndraM: 🤔 I see this static constructor as for "default" service, so it should not have any configuration IMHO. Maybe rather automatic free port detection should be added.

However if you want to override this default port settings, you can use the default constructor and set driver arguments as needed.

PR #895, opened 2021-04-22
Fix browser capabilities to be set to driver instance when using ChromeDriver

OndraM:
As part of #892 I noticed capabilities received from the browser are not set to instance of `ChromeDriver`, when the driver is created using `ChromeDriver::start()` and we receive response to the new session command.

Instead, the driver was created with capabilities used to create the browser, which is wrong and also different behavior than is when using `RemoteWebDriver::create()`.

I also added `startUsingDriverService()` static constructor method, to make it consistent with FirefoxDriver.

### DIFF
--- a/tests/functional/Chrome/ChromeDriverTest.php
+++ b/tests/functional/Chrome/ChromeDriverTest.php
@@ -36,9 +36,12 @@ class ChromeDriverTest extends TestCase
     public function testShouldStartChromeDriver()
     {
         $this->startChromeDriver();
-
         $this->assertInstanceOf(ChromeDriver::class, $this->driver);
         $this->assertInstanceOf(DriverCommandExecutor::class, $this->driver->getCommandExecutor());
+
+        // Make sure actual browser capabilities were set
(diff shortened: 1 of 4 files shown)


tests/functional/Chrome/ChromeDriverTest.php
```diff
@@ -33,18 +33,36 @@ protected function tearDown(): void
...
     {
-        $this->startChromeDriver();
-
+        $this->startChromeDriver($isW3cDialect);
         $this->assertInstanceOf(ChromeDriver::class, $this->driver);
         $this->assertInstanceOf(DriverCommandExecutor::class, $this->driver->getCommandExecutor());
 
+        // Make sure actual browser capabilities were set
+        $this->assertNotEmpty($this->driver->getCapabilities()->getVersion());
+        $this->assertNotEmpty($this->driver->getCapabilities()->getCapability('goog:chromeOptions'));
```
OndraM, line 48: This shows the fix - this would be failing before. `$this->driver->getCapabilities()` was containing only the capabilities used for creation, not the actual ones from the browser.

williamdes (approved):
LGTM

PR #914, opened 2021-06-23
Add aliases for new vendor namespace

mauriciofauth:
Adds `PhpWebDriver` as an alias for `Facebook\WebDriver`, except for `@deprecated` classes.

This allows changing the vendor namespace without a backwards compatibility break.

This is similar to how Twig handled the namespace change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,8 @@
   },
   "autoload": {
     "psr-4": {
-      "Facebook\\WebDriver\\": "lib/"
+      "Facebook\\WebDriver\\": "lib/",
+      "PhpWebDriver\\": "src/"
     },
     "files": [
       "lib/Exception/TimeoutException.php"
--- a/lib/AbstractWebDriverCheckboxOrRadio.php
+++ b/lib/AbstractWebDriverCheckboxOrRadio.php
@@ -238,3 +238,5 @@ abstract class AbstractWebDriverCheckboxOrRadio implements WebDriverSelectInterf
         }
     }
 }
+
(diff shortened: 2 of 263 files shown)


andrewnicols:
I think that this change needs much more thought, including:
1. the actual namespace to be used
2. other related areas which need to be updated
3. the length of any deprecation
4. which version this is applied to

My preference would be to make the namespace something like `PhpWebDriver\WebDriver` because `PhpWebDriver` is the vendor, and `WebDriver` is the project. I'm not sure what @OndraM thinks about this?

If we are to start such a rename the we should make the change canonical and provide the `Facebook\WebDriver` namespace for backwards compatability rather than providing a future-compatible `PhpWebDriver` namespace. That is to say that we should:
a) Update the namespace declaration
b) Swap the `class_alias()` around
c) Use the `::class` constant

I also feel that this set of commits should include all documentation updates, changes to examples, and more.

This will clearly need a new major version too so we should defer making such a change until we're ready to release a new major.

mauriciofauth:
I opened a new pull request (https://github.com/php-webdriver/php-webdriver/pull/915) with the requested changes.

> This will clearly need a new major version too so we should defer making such a change until we're ready to release a new major.

I don't think this need to be in a new major version, since it has no backwards compatibility breaks. It can be added to `1.12.0` for example. With the `@deprecated` annotations added in https://github.com/php-webdriver/php-webdriver/pull/915, it will help users to easily change to the new namespace without issues. Then the `Facebook` vendor namespace could be removed in `2.0.0`.

andrewnicols:
Why open a new pull request and lose the discussion history?

PR #915, opened 2021-06-23
Add new vendor namespace

MauricioFauth:
- Renames the `Facebook` vendor namespace to `PhpWebDriver` (including classes with `@deprecated` annotation).
- Adds class aliases for `Facebook` vendor namespace.
- Adds `@deprecated` annotation for classes with `Facebook` vendor namespace.

Related to https://github.com/php-webdriver/php-webdriver/pull/914.

The new files in the `src` directory are used by Composer to generate its class maps. And the fully qualified `\false` is required in PHP 5 to ensure it is a dead code, since it can be overwritten in PHP 5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ $serverUrl = 'http://localhost:4444';
 Now you can start browser of your choice:
 
 ```php
-use Facebook\WebDriver\Remote\RemoteWebDriver;
+use PhpWebDriver\WebDriver\Remote\RemoteWebDriver;
 
 // Chrome
 $driver = RemoteWebDriver::create($serverUrl, DesiredCapabilities::chrome());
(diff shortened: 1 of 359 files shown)


mauriciofauth:
As I mentioned earlier in https://github.com/php-webdriver/php-webdriver/pull/914:

I don't think this need to be in a new major version, since it has no backwards compatibility breaks. It can be added to `1.12.0` for example. And the `@deprecated` annotations will help users to easily change to the new namespace without issues. Then the `Facebook` vendor namespace could be safely removed in `2.0.0`. Or a hard deprecation could be added in `2.0.0` with `trigger_error` and then removed in `3.0.0`.

OndraM:
I'm closing this PR, because its not feasible to keep this branch up-to-date, its already full of conflicts... However this PR would definitely serve as an inspiration once we will actually do the namespace change. Thanks @MauricioFauth!

PR #930, opened 2021-08-24
Upgrade to php-cs-fixer 3

OndraM:
This is a prerequisite for future Symfony 6 support, as discussed in #924 .

It also separates php-cs-fixer to its own composer.json, to not interfere with php-webdriver dependencies.

cc @driesvints


driesvints:
Thanks @OndraM. This lgtm 👍

.gitignore
```diff
@@ -1,7 +1,9 @@
 composer.phar
 composer.lock
 vendor
+tools/php-cs-fixer/vendor
```
williamdes, line 4: `.gitattributes` should probably not export this folder at all

tests/unit/Firefox/FirefoxOptionsTest.php
```diff
@@ -6,7 +6,7 @@
 
 class FirefoxOptionsTest extends TestCase
 {
-    /** @var [] */
+    /** @var array */
```
williamdes, line 9:
```suggestion
    /** @var array<string,bool> */
```

williamdes (approved):
LGTM, only gitattributes needs to be adjusted

.gitignore
OndraM: Good point, fixed

PR #935, opened 2021-09-26
Fix PHP 8.1 compatibility

kylekatarnls:


lib/Chrome/ChromeOptions.php
```diff
@@ -43,6 +44,7 @@ class ChromeOptions implements JsonSerializable
      *
      * @return array
      */
+    #[ReturnTypeWillChange]
```
williamdes, line 47:
```suggestion
    #[\ReturnTypeWillChange]
```
This avoids a use

williamdes (approved):

lib/Chrome/ChromeOptions.php
kylekatarnls: I don't think it's whished to hide classes coupling and it's inconsistent with what is done for `JsonSerializable`.

driesvints: We're also using `#[\ReturnTypeWillChange]` everywhere in Laravel as does Symfony.

driesvints (approved):

lib/Chrome/ChromeOptions.php
kylekatarnls: OK but is there any reason to use FCQN for ReturnTypeWillChange and not for JsonSerializable, we could also `implements \JsonSerializable` and "save" an import, but using FCQN for some but not all sounds a bit inconsistent.

Checking a bit Laravel code, it sounds like it declares all the imports even the classes without namespace (like `stdClass`, `LogicException`, `Throwable`) and honestly I find this good as the file import section gives you the full picture of the coupling, but then why `ReturnTypeWillChange` should be handled differently?

driesvints: > but then why ReturnTypeWillChange should be handled differently?

@kylekatarnls no specific reason. Probably because it's an attribute. We just followed Symfony's example. "real" classes like your examples are imported.

OndraM:
@kylekatarnls Could you please rebase the branch against current `main`?

PR #942, opened 2021-10-24
Adding cropped screenshot

iliaamiri:
takeCroppedScreenshot is a method that allows you to take cropped screenshots from the browser.

The return value is the screenshot's image source.


OndraM:
Agree with @williamdes, this should be handled in some other library. Php-webdriver is mainly low-level implementation of WebDriver protocol for PHP (with few exceptions), so this is out of scope.

PR #951, opened 2021-11-08
Fix phpdoc for getAttribute and getDomProperty to match the specs

OndraM:
According to W3C Specs, [getElementAttribute()](https://w3c.github.io/webdriver/#get-element-attribute) will return `true`, `null` or element value (which will be string, in case of attribute).

And `getDomProperty()` could return [according to specs](https://w3c.github.io/webdriver/#get-element-property) either `null` or "value of property" which does not guarantee anything about type, so lets just use mixed.


williamdes (approved):
💯

OndraM:
The failed unit tests build on PHP 5.6 is unrelated, see https://github.com/php-coveralls/php-coveralls/issues/323.

PR #958, opened 2022-01-06
Fix PHP 8.1 deprecations in the Cookie class

jvdlaar:
This throws deprecation messages when running on PHP 8.1. Because this library also supports very old php versions I guess we cannot just set return types.

### DIFF
--- a/lib/Cookie.php
+++ b/lib/Cookie.php
@@ -207,16 +207,19 @@ class Cookie implements \ArrayAccess
         return $cookie;
     }
 
+    #[\ReturnTypeWillChange]
     public function offsetExists($offset)
     {
         return isset($this->cookie[$offset]);
     }
 
+    #[\ReturnTypeWillChange]
     public function offsetGet($offset)
     {
         return $this->offsetExists($offset) ? $this->cookie[$offset] : null;
     }
 
+    #[\ReturnTypeWillChange]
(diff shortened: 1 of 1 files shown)


williamdes (approved):

tomsykes:
I'm seeing deprecation warnings for this file in symfony 5.4, on PHP 7.4, and these changes don't fix them.
Adding a PHPDoc return type to these ArrayAccess functions should fix for both

dunglas (approved):

jmalinens:
I have the same issue. without `#[\ReturnTypeWillChange]` I get:

```
Fatal error: During inheritance of ArrayAccess: Uncaught Behat\Testwork\Call\Exception\CallErrorException: 8192: Return type of Facebook\WebDriver\Cookie::offsetExists($offset) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

dsuurlant:
We'd love to have these changes in a new version -- any idea when 1.13.0 will be released?

OndraM:
Hi, this was just released in version 1.12.1, [see changelog](https://github.com/php-webdriver/php-webdriver/blob/main/CHANGELOG.md#1121---2022-05-03).

PR #966, opened 2022-02-10
proxy

tomazov:
advanced request settings

### DIFF
--- a/lib/Remote/HttpCommandExecutor.php
+++ b/lib/Remote/HttpCommandExecutor.php
@@ -182,10 +182,19 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
      * @param string|null $http_proxy
      * @param int|null $http_proxy_port
      */
-    public function __construct($url, $http_proxy = null, $http_proxy_port = null)
+    public function __construct($url, $http_proxy = null, $http_proxy_port = null, $http_proxy_user = null, $http_proxy_pass = null)
     {
         self::$w3cCompliantCommands = array_merge(self::$commands, self::$w3cCompliantCommands);
 
+        $uri = parse_url($http_proxy);
+        if ($uri !== false) {
+            $http_proxy = $uri['host'];
+            $http_proxy_port = $uri['port'];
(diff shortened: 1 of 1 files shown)


tomazov:
advanced request settings

lib/Remote/HttpCommandExecutor.php
```diff
@@ -182,10 +182,19 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
...
         self::$w3cCompliantCommands = array_merge(self::$commands, self::$w3cCompliantCommands);
 
+        $uri = parse_url($http_proxy);
+        if ($uri !== false) {
+            $http_proxy = $uri['host'];
+            $http_proxy_port = $uri['port'];
+            $http_proxy_user = $uri['user'];
+            $http_proxy_pass = $uri['pass'];
+        }
+        unsewt($uri);
```
williamdes, line 196:
```suggestion
        unset($uri);
```

lib/Remote/HttpCommandExecutor.php
```diff
@@ -182,10 +182,19 @@ class HttpCommandExecutor implements WebDriverCommandExecutor
      * @param string|null $http_proxy
      * @param int|null $http_proxy_port
```
williamdes, line 183:
```suggestion
     * @param int|null $http_proxy_port
     * @param string|null $http_proxy_user
     * @param string|null $http_proxy_pass
```

lib/Remote/HttpCommandExecutor.php
OndraM, line 196: I would remove this whole part of parsing URI. If someone needs this (which would be very uncommon case), they can do this parsing themselves and provide the values directly.

OndraM (changes requested):
While I can imagine there is possibility someone would run remote end behind a proxy, this PR is not even lintable... @tomazov Are you willing to finish the PR? If so, please:

1) Rebase the branch with current main branch
2) Address the comments in PR

Thanks!

PR #983, opened 2022-06-09
Fix PHP 8.1 deprecation in RemoteWebDriver::takeScreenshot

### LINKED ISSUE #969: Passing `null` to `base64_decode` is deprecated
### Bug description
https://github.com/phpmyadmin/phpmyadmin/runs/5276751033?check_suite_focus=true#step:15:29
```
PHP Deprecated:  base64_decode(): Passing null to parameter #1 ($string) of type string is deprecated in /home/runner/work/phpmyadmin/phpmyadmin/vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php on line 367
```

### Details

* Php-webdriver version: 1.12.0
* PHP version:  8.1.2

MauricioFauth:
Passing `null` to `base64_decode` on PHP 8.1 triggers a [deprecation](https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg) message.

`base64_decode` returns an empty string when `null` is passed as a parameter. For the next major version of php-webdriver, I think it's better to throw an exception instead of returning an empty string.

- Fixes #969.

### DIFF
--- a/lib/Remote/RemoteWebDriver.php
+++ b/lib/Remote/RemoteWebDriver.php
@@ -364,6 +364,10 @@ class RemoteWebDriver implements WebDriver, JavaScriptExecutor, WebDriverHasInpu
      */
     public function takeScreenshot($save_as = null)
     {
+        if ($this->executor === null) {
+            return '';
+        }
+
         $screenshot = base64_decode($this->execute(DriverCommand::SCREENSHOT), true);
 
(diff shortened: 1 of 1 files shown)


williamdes (approved):

OndraM:
Replaced by #996 - have a look there please. Thanks for your help!

Issue #986, opened 2022-06-20
createBySessionID() and getAllSessions() do not work anymore?


RestOp:
Hello,

### Bug description
Q1: When i try to use createBySessionID() i get no action and no error even if i use unreal/random sessionID.

Q2: When i try to use getAllSessions() i got

`<b>Fatal error</b>: Uncaught Facebook\WebDriver\Exception\UnknownCommandException: Unable to find handler for (GET) /sessions in /usr/share/nginx/html/vendor/php-webdriver/webdriver/lib/Exception/WebDriverException.php:137 Stack trace: #0 /usr/share/nginx/html/vendor/php-webdriver/webdriver/lib/Remote/HttpCommandExecutor.php(372): Facebook\WebDriver\Exception\WebDriverException::throwException('unknown command', 'Unable to find ...', Array) #1 /usr/share/nginx/html/vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php(572): Facebook\WebDriver\Remote\HttpCommandExecutor-&gt;execute(Object(Facebook\WebDriver\Remote\WebDriverCommand)) #2 /usr/share/nginx/html/test.php(58): Facebook\WebDriver\Remote\RemoteWebDriver::getAllSessions('http://localhost....') #3 {main} thrown in <b>/usr/share/nginx/html/vendor/php-webdriver/webdriver/lib/Exception/WebDriverException.php</b> on line <b>137</b><br />`

### How could the issue be reproduced
Start some Chrome sessions (or do not start - result is the same), run the code.

### Code:
```php
**Q1:**
namespace Facebook\WebDriver;
use Facebook\WebDriver\Remote\RemoteWebDriver;
use Facebook\WebDriver\Chrome\ChromeOptions;
use Facebook\WebDriver\Remote\DesiredCapabilities;
require_once('vendor/autoload.php');
$serverUrl = 'http://localhost:4444/';
$id = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx";
$driver = RemoteWebDriver::createBySessionID($id, $serverUrl);

**Q2:**
namespace Facebook\WebDriver;
use Facebook\WebDriver\Remote\RemoteWebDriver;
use Facebook\WebDriver\Chrome\ChromeOptions;
use Facebook\WebDriver\Remote\DesiredCapabilities;
require_once('vendor/autoload.php');
$serverUrl = 'http://localhost:4444/';
[...]

PR #994, opened 2022-08-23
Mark getAllSessions() deprecated in W3C WebDriver

OndraM:
See issue in #986 . The `GET /sessions` endpoint is not part of W3C WebDriver, because of security reasons (so that you cannot get and steal other session id).

Although some remote ends (chromedriver) implement it, it is no longer part of the standard. If someone want to call the command eg. on Chromedriver without deprecation warnings (but on its own responsibility), it could still be achieved using [custom commands](https://github.com/php-webdriver/php-webdriver/wiki/Custom-commands).

```php
$sessions = $driver->executeCustomCommand(
    '/sessions',
    'GET'
);

// ...
```


OndraM:
Test comment.

PR #995, opened 2022-08-23
Replace no-response probot with 'lee-dohm/no-response' action

OndraM:
[probot/no-response](https://github.com/probot/no-response) is deprecated and no longer working.


OndraM:
Test issue comment.

williamdes (commented):
Thanks

PR #996, opened 2022-08-25
Separate takeScreenshot logic to better handle errors and to fix PHP 8.1 deprecations

OndraM:
Apart from separating screenshot/filesystem logic from two places, this should better handle errors and solve #969 and #983.

The previous behavior was to silently ignore if screenshot command returned empty data. This will now throw an `WebDriverException`.

### DIFF
--- a/lib/Support/ScreenshotHelper.php
+++ b/lib/Support/ScreenshotHelper.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Facebook\WebDriver\Support;
+
+use Facebook\WebDriver\Exception\WebDriverException;
+use Facebook\WebDriver\Remote\DriverCommand;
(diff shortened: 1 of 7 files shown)


lib/Support/ScreenshotHelper.php
```diff
@@ -0,0 +1,77 @@
...
+    /** @var RemoteExecuteMethod */
+    private $executor;
+
+    public function __construct(RemoteExecuteMethod $executor)
+    {
+        $this->executor = $executor;
+    }
+
+    /**
+     * @param ?string $saveAs
```
williamdes, line 23:
```suggestion
     * @param null|string $saveAs
```

williamdes (approved):

MauricioFauth (approved):

PR #945, opened 2021-11-02
Throw NoSuchElementException instead of potential fatal error

bchecketts:
Fix for #944 

Throw an `Facebook\WebDriver\Exception\NoSuchElementException` instead of causing fatal error.

PR #889, opened 2021-04-16
Replace Sami by Doctum and re-build Docs

williamdes:

PR #1001, opened 2022-09-01
Throw exception instead of fatal error on invalid server response

### LINKED ISSUE #944: PHP Fatal Error in RemoteWebDriver::findElement()
### Bug description

RemoteWebDriver::execute() may return a null value.  When this happens from `findElement()`, it causes `JsonWireCompat::getElement()` to throw a PHP Fatal error

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Facebook\WebDriver\Remote\JsonWireCompat::getElement() must be of the type array, null given, called in <myPath>/vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php on line 215 and defined in <myPath>/api/vendor/php-webdriver/webdriver/lib/Remote/JsonWireCompat.php:21
Stack trace:
#0 <myPath>/vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php(215): Facebook\WebDriver\Remote\JsonWireCompat::getElement()
#1 <myfile>.php(410): Facebook\WebDriver\Remote\RemoteWebDriver->findElement()
```

### How could the issue be reproduced
I wish I could capture some details about the page content, but the PHP Fatal error causes the session to die immediately, so I can't see what is happening.

* Php-webdriver version: 1.11.0
* PHP version:  7.2.24
* Browser driver (chromedriver/geckodriver...) version:  95.0.4638.69
* Browser used + version: 
* Operating system:

OndraM:
In some hard-to-reproduce scenarios, remote end returns response, however, it contains `null` in value (or does not contain value at all). This causes php-webdriver to fail on fatal/type error. 

Instead, `UnknownErrorException` will be throw, to allow better debugging.

Fixes #944, replaces #945

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ This project versioning adheres to [Semantic Versioning](http://semver.org/).
 - Handle errors when taking screenshots. `WebDriverException` is thrown if WebDriver returns empty or invalid screenshot data.
 - Deprecate `FirefoxDriver::PROFILE` constant. Instead, use `setProfile()` method of `FirefoxOptions` to set Firefox Profile.
 
+### Fixed
+- Throw `UnknownErrorException` instead of fatal error if remote end returns invalid response for `findElement()`/`findElements()` commands.
+
 ## 1.12.1 - 2022-05-03
 ### Fixed
 - Improper PHP documentation for `getAttribute()` and `getDomProperty()`.
--- a/lib/Remote/RemoteWebDriver.php
+++ b/lib/Remote/RemoteWebDriver.php
@@ -2,6 +2,7 @@
 
 namespace Facebook\WebDriver\Remote;
 
+use Facebook\WebDriver\Exception\UnknownErrorException;
 use Facebook\WebDriver\Interactions\WebDriverActions;
 use Facebook\WebDriver\JavaScriptExecutor;
 use Facebook\WebDriver\Support\ScreenshotHelper;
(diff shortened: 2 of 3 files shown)


williamdes (approved):
👌🏻 

Could you also have a look to the lint docs complaining about `createBySessionID` ?

OndraM:
> Could you also have a look to the lint docs complaining about createBySessionID ?

It appeared after we started to use Doctum. I mentioned it here: https://github.com/php-webdriver/php-webdriver/pull/889#issuecomment-822067881

And to be honest, I'd really like to get rid of there errors, because they are filling the pull requests with unexpected and confusing noise... Any ideas?

OndraM:
@williamdes 

> Could you also have a look to the lint docs complaining about `createBySessionID` ?

It appeared after we started to use Doctum. I mentioned it here: https://github.com/php-webdriver/php-webdriver/pull/889#issuecomment-822067881

And to be honest, I'd really like to get rid of there errors, because they are filling the pull requests with unexpected and confusing noise... Any ideas?

williamdes:
> @williamdes
> 
> > Could you also have a look to the lint docs complaining about `createBySessionID` ?
> 
> It appeared after we started to use Doctum. I mentioned it here: [#889 (comment)](https://github.com/php-webdriver/php-webdriver/pull/889#issuecomment-822067881)
> 
> And to be honest, I'd really like to get rid of there errors, because they are filling the pull requests with unexpected and confusing noise... Any ideas?

I think a setting could be added to Doctum for functions having too much parameters, it would silence this issue

PR #1004, opened 2022-09-04
Minor project fixes and improvements

williamdes:


.github/workflows/docs-publish.yml
```diff
@@ -1,5 +1,8 @@
 name: Publish API documentation
 
+permissions:
+  contents: read
```
OndraM, line 4: Is this sufficient? This workflow pushes to the gh-pages branch.

williamdes: I think so because we push from a ssh key and not from the token obtained on the running workflow

OndraM: Ok, we will see :)

OndraM:
Thanks @williamdes!

PR #1006, opened 2022-09-07
Update readme, as the W3C WebDriver support is not experimental

OndraM:
The W3C DebDriver support in php-webdriver is probably better then support for JsonWire ever was, so its unnecessary to describe it as "experimental".

Also fixed few http links


williamdes (commented):
Thanks

PR #1009, opened 2022-09-08
Do not submit coveralls on pull_request event

OndraM:
Apparently, we are impacted by https://github.com/lemurheavy/coveralls-public/issues/1653 , leading to coverage decrease report for *every* pull request (the reported descrease is from 65 % to to 35 %, even when no code changes).

This is attempt for a workaround - do not submit coverage reports for the `pull_request` events, but only for `push`.


OndraM:
OTOH, this means no coverage reports for pull request from forks. :man_facepalming:

PR #1026, opened 2022-10-13
Update sauce connect action

OndraM:
> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: saucelabs/sauce-connect-action, saucelabs/sauce-connect-action


OndraM:
Ok, it works, however the action itself has Node16 already merged, but not released: https://github.com/saucelabs/sauce-connect-action

So as we now use  `@v2` we will get the Node16 automatically once they tag new version (if it will be tagged as 2.x). So lets merge this for now.

PR #1027, opened 2022-10-13
Refactor internal exceptions

OndraM:
Several Exception-refactoring issues are included in this PR:
- Exceptions based on W3C WebDriver/JsonWire protocol are kept in `Facebook\WebDriver\Exception`, however, all internal exceptions were separated to Facebook\WebDriver\Exception\Internal .
- Both protocol-based and internal exceptions implements new `PhpWebDriverExceptionInterface`, which allows to be catched by some higher-level library using php-webdriver.
- There should be no PHP SPL or other root namespace exceptions thrown from php-webdriver, all of them should be using internal exceptions (like `Facebook\WebDriver\Exception\Internal\LogicException`, `IOException` etc.), some of them were newly added
- Most of exception throwing was refactored to use more descriptive static constructors instead of `new ...Exception`.


williamdes (approved):

williamdes (approved):

inxilpro:
Hi @OndraM — we just ran into a bug related to exceptions moving to the `Internal` namespace between 1.13 and 1.14. We were previously catching `Facebook\WebDriver\Exception\WebDriverCurlException`, but it has been renamed and deprecated. (Renaming an exception feels like a breaking change, fwiw.)

Is it safe to use `UnexpectedResponseException`, or are exceptions in the `Internal` namespace not meant to be used outside of the package?

Issue #1022, opened 2022-10-07
Release 1.13 causes RemoteWebdriver Wait Until error (RemoteWebDriver.php:581)


gwillyoo:
### Bug description
Make webdriver wait until expected condition. But suddenly shows error. Here is the stack.

```php
vendor/php-webdriver/webdriver/lib/WebDriverWait.php(54): call_user_func()

[internal function]: Facebook\WebDriver\WebDriverExpectedCondition::Facebook\WebDriver\{closure}()

vendor/php-webdriver/webdriver/lib/WebDriverExpectedCondition.php(178): Facebook\WebDriver\Remote\RemoteWebElement->isDisplayed()

vendor/php-webdriver/webdriver/lib/Remote/RemoteWebElement.php(345): Facebook\WebDriver\Remote\RemoteExecuteMethod->execute()


vendor/php-webdriver/webdriver/lib/Remote/RemoteExecuteMethod.php(27): Facebook\WebDriver\Remote\RemoteWebDriver->execute()

Call to a member function getBrowserName() on null in /home/forge/bigman.app/vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php:581
```

### How could the issue be reproduced

using Elgalu/Selenium

```php browser()->wait()->until(WebDriverExpectedCondition::visibilityOfElementLocated(WebDriverBy::id('withdrawal-pending-table')))```

**Comment by gwillyoo** [comment]
Yes, it is. I keep the browser session.

**Comment by OndraM** [comment]
Hi @gwillyoo, this is fixed in just released version 1.13.1.

Thanks for reporting this!

**Comment by y2wde** [comment]
> OndraM

I updated php-webdriver last week from 1.11.1 to 1.13.1 and I can confirm that wait-until now runs into an error, if the element not exists. 

For example:

```
$this->driver->wait(10, 500)->until(
 WebDriverExpectedCondition::presenceOfElementLocated(WebDriverBy::id('shopping-cart-box'))
);
```

So this bug still exists and is not fixed with 1.13.1.

PR #1032, opened 2022-10-21
Improve session reusing, require capabilities to be set

OndraM:
This is a proper fix for issues like #1022 etc.

When RemoteWebDriver instance was created using `createBySessionID()`, it was missing value of capabilities (it was null). This lead to various unexpected issues, for example with instance crated like this, you are not able to get the name of the running browser. Which is unexpected, not type-safe and leading to errors.

With this PR, we try to read capabilities from Selenium Grid server (there is no endpoint defined in W3C WebDriver, it is only the API of Selenium Server itself). Alternatively, if you don't use Selenium Grid, you must explicitly provide capabilities using the sixth parameter of `createBySessionID()`, see example. Thank to this, you can always be sure `$driver->getCapabilities()`, `$driver->getCapabilities()->getBrowserName()` etc. will return a sane value.

```php
$originalDriver = RemoteWebDriver::create($serverUrl, $desiredCapabilities);
$sessionId = $originalDriver->getSessionID(); // save sessionId to variable
$originalCapabilities = $originalDriver->getCapabilities(); // save capabilities to variable

// reuse driver using the session ID and capabilities
$driverFromSessionId = RemoteWebDriver::createBySessionID($sessionId, $serverUrl, null, true, $originalCapabilities);
```

If you don't use Selenium Grid and you don't provide capabilities, you won't be able to crate the RemoteWebDriver instance using `createBySessionID()` method.

---

- [x] Add CHANGELOG entry

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -162,6 +162,7 @@ jobs:
           CHROMEDRIVER_PATH: "${{ (matrix.browser == 'chrome' && !matrix.selenium-server) && '/usr/local/share/chrome_driver/chromedriver' || '' }}"
           GECKODRIVER_PATH: "${{ (matrix.browser == 'firefox' && !matrix.selenium-server) && '/usr/local/share/gecko_driver/geckodriver' || '' }}"
           DISABLE_W3C_PROTOCOL: "${{ matrix.w3c && '0' || '1' }}"
+          SELENIUM_SERVER: "${{ matrix.selenium-server && '1' || '0' }}"
         run: |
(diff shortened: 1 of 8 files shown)


williamdes (approved):

williamdes (approved):

gwillyoo:
Hi @OndraM , can you update this code :

`$existingCapabilities = func_num_args() > 5 ? func_get_arg(5) : null;`


When I want to reusing the session because of that function i should create the session like this :

`RemoteWebDriver::createBySessionID($this->session_id, $this->host, null, true, $this->capabilities, $this->capabilities)`

Please advise. Thank you 🙏🏼

PR #1045, opened 2023-01-31
Add typehints where safe, remove superfluous phpdoc

OndraM:
- Add return types to test methods
- Add typehints to test methods
- Remove superfluous phpdoc ( = phpdoc, which adds no information on top of what is already in typehint ) in the whole codebase (lib and tests)

Typehints were not added to the source code except for tests because this could cause BC break. (This is also why the rule is commented in `.php-cs-fixer.dist.php`)


williamdes (approved):
If type hints are there and phpdoc does not add a description for the type then yeah remove them

PR #1055, opened 2023-03-06
add executeCdp function

liujianchun:
Add function executeCdp() to support execute 'Chrome DevTools Protocol'. https://chromedevtools.github.io/devtools-protocol/tot/Browser/


OndraM:
Hi @liujianchun , thanks for the pull request!

However, I believe we already have this implemented. See wiki:
https://github.com/php-webdriver/php-webdriver/wiki/Chrome#chrome-devtools-protocol-cdp

Issue #1062, opened 2023-03-27
Argument 1 passed to Facebook\WebDriver\Remote\JsonWireCompat::getElement() must be of the type array, string given


Mike-Benoit:
### Bug description

I'm seeing this type error exception only in some cases, but it seems outside of my control, here is the full trace:

```
Error: Argument 1 passed to Facebook\WebDriver\Remote\JsonWireCompat::getElement() must be of the type array, string given, called in ./vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php on line 226
#0 ./vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php(226): Facebook\WebDriver\Remote\JsonWireCompat::getElement()
#1 ./vendor/php-webdriver/webdriver/lib/WebDriverExpectedCondition.php(140): Facebook\WebDriver\Remote\RemoteWebDriver->findElement()
#2 [internal function]: Facebook\WebDriver\WebDriverExpectedCondition::Facebook\WebDriver\{closure}()
#3 ./vendor/php-webdriver/webdriver/lib/WebDriverWait.php(54): call_user_func()
#4 ./selenium.php(1870): Facebook\WebDriver\WebDriverWait->until()
#5 ./selenium.php(2496): Facebook\WebDriver\Selenium->navigateToHome()
```

### How could the issue be reproduced

Although I can't reproduce it consistently, it occurs about 60% of the time. The code is quite simple:

```php
$this->driver->wait( 60, 500 )->until( WebDriverExpectedCondition::presenceOfElementLocated( WebDriverBy::xpath( '//nav/ul/li/*[contains( text(), "Home")]' ) ) );
```

### Expected behavior

No PHP exception to be triggered.

### Php-webdriver version

1.14.0

### PHP version

7.4.3

### How do you start the browser driver or Selenium server

chromedriver

### Selenium server / Selenium Docker image version

4.8.1

### Browser driver (chromedriver/geckodriver...) version

111.0.5563.64

### Browser name and version

Chrome

### Operating system

Ubuntu 18.04

Mike-Benoit:
I managed to get some additional information:

var_dump( WebDriverBy::xpath( '//nav/ul/li/*[contains( text(), "Home")]' ) )
```
object(Facebook\WebDriver\WebDriverBy)#40 (2) {
  ["mechanism":"Facebook\WebDriver\WebDriverBy":private]=>
  string(5) "xpath"
[...]

PR #1064, opened 2023-04-25
Do not fail on fatal error on unexpected response when getting element(s)

OndraM:
There were various reports (#944, #1062) when some unexpected response from selenium causes php-webdriver to fail on fatal error because of types incompatibility (we expect array but selenium returns some other error).

This makes sure only array is passed to the `getElement` method.


williamdes (approved):

OndraM:
> (I'm still considering whether not to change type of `JsonWireCompat::getElement` to mixed and make the assertion/exception throwing inside...)

Yes, i reworked the PR so the response content assertion is done in one place, not on six different places.

williamdes (approved):

PR #1067, opened 2023-07-12
Update WebDriverSelect test to support multiple selectors

oleg-andreyev:
This commit updates WebDriverSelectTest.php to provide support for selecting multiple selectors. 
It extends the testShouldCreateNewInstanceForSelectElementAndDetectIfItIsMultiple() function with a data provider to test multiple select elements.
The changes also include corrections to WebDriverSelect.php to properly adhere to spec for Boolean attributes, by ensuring that the attribute value is 'true' or 'multiple'. 
Additional select options have also been added to the form.html file to support the new tests.

https://html.spec.whatwg.org/#boolean-attributes

### DIFF
--- a/lib/WebDriverSelect.php
+++ b/lib/WebDriverSelect.php
@@ -26,7 +26,14 @@ class WebDriverSelect implements WebDriverSelectInterface
         }
         $this->element = $element;
         $value = $element->getAttribute('multiple');
-        $this->isMulti = $value === 'true';
+
+        /**
+         * There is a bug in safari webdriver that returns 'multiple' instead of 'true' which does not match the spec.
+         * Apple Feedback #FB12760673
+         *
+         * @see https://www.w3.org/TR/webdriver2/#get-element-attribute
+         */
+        $this->isMulti = $value === 'true' || $value === 'multiple';
     }
 
     public function isMultiple()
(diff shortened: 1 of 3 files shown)


OndraM (commented):
Hi @oleg-andreyev - how did you came across this issue? Because if I'm not mistaken, this change should not be necessary and this should already work without an issue.

While in DOM it is true that boolean attributes could be either "true", empty or have the attribute's canonical name, according [WebDriver protocol specs](https://www.w3.org/TR/webdriver2/#get-element-attribute), boolean attributes are always resolved to "true" when reading their value using WebDriver.  So the value is always "normalized" for us to `true`.

I also tried it just now - if you remove the `|| $value === 'multiple';` condition, it still works and the tests pass, which confirms that this is already handled in WebDriver.

OndraM (changes requested):
So I tried to removed the change but kept the modified tests and run the tests. This is the result: https://github.com/php-webdriver/php-webdriver/actions/runs/5657268611

... so its broken only in Safari, meaning this is some kind of Safardriver bug, because this behavior clearly not adhere to W3C WebDriver standard.

I'm ok merging this change to improve cross-browser compatibility, but could you please modify the comment to clearly state this modified if condition is there only because of this bug in Safaridriver? Great would be to link some Safari bugreport or some other reference to this behavior, if you could find it anywhere.

oleg-andreyev:
@OndraM updated comment.

OndraM:
Merging, thanks!

(The broken Chromedriver test are unrelated, see #1069.)

PR #1069, opened 2023-07-27
Fix path to chromedriver in tests on latest Ubuntu

OndraM:
The path [is now different](https://github.com/actions/runner-images/blob/ubuntu22/20230724.1/images/linux/Ubuntu2204-Readme.md#environment-variables-1) and the tests are now failing with
`Facebook\WebDriver\Exception\Internal\IOException: File is not executable. Make sure the path is correct or use environment variable to specify location of the executable. ("/usr/local/share/chrome_driver/chromedriver")`

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -160,7 +160,7 @@ jobs:
           # When running directly against Firefox, we test only using geckodriver (not against legacy Firefox =<45), so we must declare GECKODRIVER=1
           GECKODRIVER: "${{ (matrix.browser == 'firefox' && !matrix.selenium-server) && '1' || '0' }}"
           # Provide CHROMEDRIVER_PATH and GECKODRIVER_PATH so that tests for local web drivers are able to start the browser
-          CHROMEDRIVER_PATH: "${{ (matrix.browser == 'chrome' && !matrix.selenium-server) && '/usr/local/share/chrome_driver/chromedriver' || '' }}"
+          CHROMEDRIVER_PATH: "${{ (matrix.browser == 'chrome' && !matrix.selenium-server) && '/usr/local/share/chromedriver-linux64/chromedriver' || '' }}"
           GECKODRIVER_PATH: "${{ (matrix.browser == 'firefox' && !matrix.selenium-server) && '/usr/local/share/gecko_driver/geckodriver' || '' }}"
           DISABLE_W3C_PROTOCOL: "${{ matrix.w3c && '0' || '1' }}"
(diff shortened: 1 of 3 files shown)


williamdes (approved):

.github/workflows/tests.yaml
```diff
@@ -160,7 +160,7 @@ jobs:
           # When running directly against Firefox, we test only using geckodriver (not against legacy Firefox =<45), so we must declare GECKODRIVER=1
           GECKODRIVER: "${{ (matrix.browser == 'firefox' && !matrix.selenium-server) && '1' || '0' }}"
           # Provide CHROMEDRIVER_PATH and GECKODRIVER_PATH so that tests for local web drivers are able to start the browser
-          CHROMEDRIVER_PATH: "${{ (matrix.browser == 'chrome' && !matrix.selenium-server) && '/usr/local/share/chrome_driver/chromedriver' || '' }}"
+          CHROMEDRIVER_PATH: "${{ (matrix.browser == 'chrome' && !matrix.selenium-server) && '/usr/local/share/chromedriver-linux64/chromedriver' || '' }}"
```
oleg-andreyev, line 163: can we use `command chromedriver` or `which chromedriver`?

we can run separate step, detect location of bin and use it in another step like this.

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

oleg-andreyev (changes requested):

.github/workflows/tests.yaml
OndraM: Yes, but this would mean rewriting this whole section to `run` script, as you unfortunately cannot use values of other environment variables inside `env` section. So even if we save the path to environment variable using `which`, it would not be possible to use the value here...

lib/Local/LocalWebDriver.php
```diff
@@ -38,8 +33,6 @@ public static function create(
     /**
      * @param string $session_id
      * @param string $selenium_server_url
-     * @param null $connection_timeout_in_ms
-     * @param null $request_timeout_in_ms
```
williamdes, line 42: why are they removed?

OndraM: Php-cs-fixer thinks they are superfluous, which - in fact - are. However this conflicts with other check, so I deleted them altogether (they should be inherited from the parent class).

williamdes: okay

williamdes (approved):

PR #1074, opened 2023-08-22
Update PHP versions in test builds, run tests primarily on PHP 8

OndraM:


.github/workflows/tests.yaml
```diff
@@ -46,11 +46,10 @@ jobs:
 
     strategy:
       matrix:
-        php-version: ['7.3', '7.4', '8.0']
+        php-version: ['7.3', '7.4', '8.0', '8.1', '8.2']
         dependencies: ['']
         include:
           - { php-version: '7.3', dependencies: '--prefer-lowest' }
-          - { php-version: '8.1', dependencies: '--ignore-platform-req=php' }
```
williamdes, line 53: You could add nightly with allowed failure

williamdes (approved):

.github/workflows/tests.yaml
OndraM: Maybe next time :)

PR #1096, opened 2024-04-24
Fix static return types in RemoteWebElement

PrinsFrank:
A couple of methods that are returning $this in the RemoteWebElement class are marked as returning an instance of RemoteWebElement itself when in fact they return a static/$this instance when the RemoteWebElement class is extended. This PR fixes those, so static analysis understands the extended class.


OndraM:
Thanks @PrinsFrank!

OndraM (approved):

PrinsFrank:
@OndraM No problem! Any chance this can be published in a new version so we can remove the PHPStan stubs from our codebase?

Issue #1107, opened 2024-08-13
Support PHP 8.4


lolli42:
### Bug description

Add PHP 8.4 to test matrix and fix "nullable" issues.

### How could the issue be reproduced

```shell
Run test suite with PHP 8.4.
```

### Php-webdriver version

1.15.1

### PHP version

8.4 alpha 4

### How do you start the browser driver or Selenium server

no need

PR #1108, opened 2024-08-13
Support PHP 8.4

lolli42:
Resolves: #1107

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -77,6 +77,10 @@ return (new PhpCsFixer\Config())
         'no_whitespace_before_comma_in_array' => ['after_heredoc' => true],
         'no_whitespace_in_blank_line' => true,
         'non_printable_character' => true,
+        'nullable_type_declaration' => [
+            'syntax' => 'question_mark',
+        ],
+        'nullable_type_declaration_for_default_null_value' => true,
         'object_operator_without_whitespace' => true,
         'ordered_class_elements' => true,
         'ordered_imports' => true,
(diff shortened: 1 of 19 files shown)


lolli42:
Fixed using the two new rules in php-cs-fixer.

lolli42:
Friendly ping.

crynobone (approved):

lolli42:
Friendly ping.
We are at PHP 8.4.0 rc2 now. Would be great to have a compat release of this library.

.php-cs-fixer.dist.php
```diff
@@ -77,6 +77,10 @@
         'no_whitespace_before_comma_in_array' => ['after_heredoc' => true],
         'no_whitespace_in_blank_line' => true,
         'non_printable_character' => true,
+        'nullable_type_declaration' => [
+            'syntax' => 'question_mark',
```
uuf6429, line 81: Just happened to see this change.

I would suggest against this, it doesn't play well with union types. E.g. `?int|float` is invalid. Just a thought.

uuf6429: Hmm, but on the other hand union types are not supported before 8.0, meanwhile 8.4 shows a deprecation message for implicit null types. 🙄

Guess this is the only way then.

OndraM (approved):
Nice, thank you!

crynobone:
@OndraM Can we have this release ASAP (as PHP 8.4 has been released)

OndraM:
Version 1.15.2 is now released.

PR #1113, opened 2024-10-30
Allow running tests in headfull mode (useful for testing)

uuf6429:
- defines new env var (`DISABLE_HEADLESS`) to specifically enable/disable headless mode
- by default headless is enabled (for BC reasons)
- documented in the same place `BROWSER_NAME` is documented

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -69,7 +69,7 @@ composer all
 composer test -- --testsuite functional
 ```
 
-If you want to run tests in different browser then "htmlunit" (Chrome or Firefox), you need to setup the browser driver (Chromedriver/Geckodriver), as it is [explained in wiki](https://github.com/php-webdriver/php-webdriver/wiki/Chrome)
+If you want to run tests in different browser then "htmlunit" (Chrome or Firefox), you need to set up the browser driver (Chromedriver/Geckodriver), as it is [explained in wiki](https://github.com/php-webdriver/php-webdriver/wiki/Chrome)
 and then the `BROWSER_NAME` environment variable:
 
 ```sh
@@ -83,5 +83,12 @@ To test with Firefox/Geckodriver, you must also set `GECKODRIVER` environment va
 ```sh
 export GECKODRIVER=1
 export BROWSER_NAME="firefox"
+composer all
+```
+
+To see the tests as they are happening (in the browser window), you can disable headless mode. This is useful eg. when debugging the tests or writing a new one:
+
+```sh
+export DISABLE_HEADLESS="1"
 composer all
 ```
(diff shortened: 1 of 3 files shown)


uuf6429:
There are two remaining classes that use headless and which I didn't change:
- https://github.com/uuf6429/php-webdriver/blob/871eeba8815e4340e158202097ad53725269cf1c/tests/functional/Chrome/ChromeDriverTest.php#L90
- https://github.com/uuf6429/php-webdriver/blob/235c2aa07f9e2686f2b876083fc081b00d330d48/tests/unit/Remote/DesiredCapabilitiesTest.php#L189 (multiple places)
Should I change them too?

.github/CONTRIBUTING.md
```diff
@@ -85,3 +85,10 @@ export GECKODRIVER=1
 export BROWSER_NAME="firefox"
 composer all
 ```
+
+To see the tests as they are happening (in the browser window), you should disable headless mode:
```
OndraM, line 89:
```suggestion
To see the tests as they are happening (in the browser window), you can disable headless mode. This is useful eg. when debugging the tests or writing a new one:
```

tests/functional/WebDriverTestCase.php
```diff
@@ -46,8 +47,12 @@ protected function setUp(): void
 
             if ($browserName === WebDriverBrowserType::CHROME) {
                 $chromeOptions = new ChromeOptions();
+
+                if (!$disableHeadless) {
```
OndraM, line 51: There is also similar case for starting a headless Firefox few lines below - this should be handled as well.

Also just a personal preference, but I would prefer moving this condition **below** the already existing `$chromeOptions->addArguments(`... call.

.github/workflows/tests.yaml
```diff
@@ -162,6 +162,7 @@ jobs:
           CHROMEDRIVER_PATH: "${{ (matrix.browser == 'chrome' && !matrix.selenium-server) && '/usr/local/share/chromedriver-linux64/chromedriver' || '' }}"
           GECKODRIVER_PATH: "${{ (matrix.browser == 'firefox' && !matrix.selenium-server) && '/usr/local/share/gecko_driver/geckodriver' || '' }}"
           DISABLE_W3C_PROTOCOL: "${{ matrix.w3c && '0' || '1' }}"
+          DISABLE_HEADLESS: "0"
```
OndraM, line 165:
```suggestion
```

As this is default behavior, I would suggest not adding this env variable for the CI run.

OndraM (changes requested):
Hi, thanks for the suggestion. It makes sense!

Could you please cover the missing case for Firefox?

.github/workflows/tests.yaml
uuf6429: I thought it would be clearer / more transparent to have it - devs would understand immediately how the pipeline behaves just by reading the file.

OndraM (approved):

OndraM:
Thanks @uuf6429!

PR #1120, opened 2025-01-01
Clarify and fix timeout value type

uuf6429:
Thanks to PHPStan, I noticed something "off" when on the client side we divide timeouts by 1000 (resulting in a float) but then this library multiplies them back by 1000 (timeouts on our side are in ms).

This library does not ensure that the correct timeout value type is passed, for example W3C does not allow float, but at the same time this library does not handle nulls.

With this PR, I'm trying to fix the following problems:
- For *JsonWire Protocol* we should allow `int|float` ("numbers"): https://www.selenium.dev/documentation/legacy/json_wire_protocol/#sessionsessionidtimeouts
- For *W3C Protocol* we should allow `null|int`: https://www.w3.org/TR/webdriver2/#timeouts
- The PHPDoc is updated as applicable

### DIFF
--- a/lib/WebDriverTimeouts.php
+++ b/lib/WebDriverTimeouts.php
@@ -28,7 +28,7 @@ class WebDriverTimeouts
     /**
      * Specify the amount of time the driver should wait when searching for an element if it is not immediately present.
      *
-     * @param int $seconds Wait time in second.
+     * @param float $seconds Wait time in seconds.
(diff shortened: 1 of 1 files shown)


OndraM:
> @OndraM small reminder about this PR. Please let me know if anything is missing or unclear!

Hi @uuf6429  and thank you, I will try to have a look next week.

PR #1122, opened 2025-02-13
Add Throwable extension to PhpWebDriverExceptionInterface

alyamovsky:
Currently, `PhpWebDriverExceptionInterface` does not extend `\Throwable`, making it difficult to catch exceptions by the interface while still accessing common exception methods like `getMessage()`, etc. This PR fixes that by extending `\Throwable`, so exceptions can be handled more easily.


williamdes (approved):

OndraM:
Thanks @alyamovsky, I will try to look on this next week.

OndraM (approved):
Thanks @alyamovsky!

PR #1124, opened 2025-03-28
Set custom webdriver port with ENV var

usernameComputer01:
Chrome: WEBDRIVER_CHROME_CUSTOM_PORT
    Firefox: WEBDRIVER_FIREFOX_CUSTOM_PORT


williamdes:
Does such env exist in another project ?
could we use more standard ones ?

usernameComputer01:
> Does such env exist in another project ? could we use more standard ones ?

I didn't find any information on the ENV standard for webdriver.

usernameComputer01:
Added checking port availability

composer.json
```diff
@@ -15,6 +15,7 @@
     "php": "^7.3 || ^8.0",
     "ext-curl": "*",
     "ext-json": "*",
+    "ext-sockets": "*",
```
williamdes, line 18: This should be in the suggested section
and the code should not require it
but maybe @OndraM has other opinions

lib/Remote/Service/DriverService.php
```diff
@@ -152,6 +152,27 @@ protected function checkWasStarted($process)
         }
     }
 
+    /**
+     * @throws RuntimeException
+     */
+    protected static function checkPortIsAvail(int $port)
+    {
+        $errTest = function ($sock) use ($port) {
```
williamdes, line 160:
```suggestion
        $errTest = static function ($sock) use ($port) {
```

composer.json
usernameComputer01: Maybe. Seems, php-cs-fixer inclueded it, because sockets used in new method

lib/Remote/Service/DriverService.php
```diff
@@ -152,6 +152,27 @@ protected function checkWasStarted($process)
         }
     }
 
+    /**
+     * @throws RuntimeException
+     */
+    protected static function checkPortIsAvail(int $port)
+    {
+        $errTest = static function ($sock) use ($port) {
```
OndraM: This method should check if the Sockets extension is enabled, and if not, it should just return true (ie. the same behavior as before - without free port detection).

OndraM (changes requested):
Thanks! Could you please update the sockets dependency? :pray:

PR #1127, opened 2025-06-02
Fix file upload endpoint for selenium 4.x

nocive:


OndraM:
Hi, I cannot commit to your branch, so I replaced this with #1143 .

Thanks @nocive!

